PR #30, opened 2014-05-11
WIXFEAT:4161 - Allow more than one MBA prereq

rseanhall:
Add the MBAPrereqSupportPackage attribute to all package types so that more than one package can be installed by the Prereq BA.

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,4 +1,4 @@
-* SeanHall: WIXFEAT:4161 - Add the MBAPrereqSupportPackage attribute to all package types so that more than one package can be installed by the Prereq BA.
+* SeanHall: WIXFEAT:4161 - Add the PrereqSupportPackage attribute to all package types so that more than one package can be installed by the Prereq BA, and the MbaPrereqPackage can be conditionally installed.
 
 * SeanHall: WIXBUG:3835 - Fix progress bug when extracting multiple packages from a container.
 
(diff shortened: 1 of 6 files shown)


raghuma4:
Hi,

 I have created my own extension schema and added custom elements to Bundle element like
```
<Bundle>
     <mySchema:Element1 attr1="value1"
                                          attr2="value2"/>
</Bundle>
```
But I would also like to add custom attribute to existing MsiPackage element in Bundle like

```
<MsiPackage  Id="p1"
                        mySchema:CustomAttr1="value" />
```
I have added custom attribute to my xsd file using 
```
<xs:attribute name="CustomAttr1" type="YesNoType">
    <xs:annotation>
      <xs:documentation>
          Documentation part
      </xs:documentation>
      <xs:appinfo>
         <xse:parent namespace="http://schemas.microsoft.com/wix/2006/wi" ref="MsiPackage" />
      </xs:appinfo>
    </xs:annotation>
  </xs:attribute>
```
As per wix documentation on MsiPackage element **Any Attribute (namespace='##other' processContents='lax') Extensibility point in the WiX XML Schema. Schema extensions can register additional attributes at this point in the schema. The extension's CompilerExtension.ParseAttribute() method will be called with the package identifier in contextValues["PackageId"].**
Reference : http://wixtoolset.org/documentation/manual/v3/xsd/wix/msipackage.html

Unfortunately I am getting error saying that **error CNDL0004: The MsiPackage element contains an unexpected attribute 'mySchema:CustomAttr1'.** The wix schema is not recognizing my custom attribute (CustomAttr1).

Please provide your suggestions on this ? Thanks in Advance

PR #132, opened 2014-08-17
Issue #4206 - Firewall port list or range not possible

NorbertHD:
Change of (i)Port variable from integer to string is necessary to handle port ranges and lists. All the other port specific code already uses strings.

### DIFF
--- a/src/ext/FirewallExtension/ca/firewall.cpp
+++ b/src/ext/FirewallExtension/ca/firewall.cpp
@@ -179,6 +179,7 @@ LExit:
     ReleaseStr(pwzCustomActionData);
     ReleaseStr(pwzName);
     ReleaseStr(pwzRemoteAddresses);
+    ReleaseStr(pwzPort);
     ReleaseStr(pwzProgram);
     ReleaseStr(pwzComponent);
     ReleaseStr(pwzDescription);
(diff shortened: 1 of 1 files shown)


robmen:
Is there no validation we should be doing? Is **any** string valid?

NorbertHD:
Any number between 1 and 65535 is valid or any range by these numbers seperated by hyphen or any combination of these seperated by comma.
Is validation necessary? I have only seen this validation where only specific words are allowed.

E.g. registry extension doesn't check the parameters.
Also the current firewall port code doesn't check if the number is between 1 and 65535.

src/ext/FirewallExtension/ca/firewall.cpp
```diff
@@ -39,7 +39,7 @@ static UINT SchedFirewallExceptions(
     LPWSTR pwzCustomActionData = NULL;
     LPWSTR pwzName = NULL;
     LPWSTR pwzRemoteAddresses = NULL;
-    int iPort = 0;
+    LPWSTR pwzPort = NULL;
```
barnson: Needs a matching call to ReleaseStr at LExit.

Have you signed an assignment agreement? We'll need that. See http://wixtoolset.org/development/assignment-agreement/ for details.

NorbertHD: I have added the missing ReleaseStr and have signed the agreement on January 9. But I have not enough WiX programming knowledge/skills to program the parameter validation.

barnson:
Thanks!

PR #180, opened 2014-11-25
WIXBUG:4609 - Fix set then get on initialized Hidden variables.

firegiantco:
Fix the handling of encrypted variables (a Bundle variable marked hidden) so that the variable can be updated.


barnson:
Resolved in WiX v3.9 R2.

PR #191, opened 2015-01-02
WIXBUG4187 - Melt changes to extract Binary table contents

stunney:
WIXBUG4187

Melt now outputs all items from the Binary table with filenames that match the `Name` value in the table.
I decided to add extraction of the binary files inline with the Extract call.
The BinaryDirectory property doesn't need to be public either but you never know.

### DIFF
--- a/src/DTF/Libraries/WindowsInstaller.Package/InstallPackage.cs
+++ b/src/DTF/Libraries/WindowsInstaller.Package/InstallPackage.cs
@@ -191,6 +191,7 @@ public class InstallPackage : Database
     {
         this.ProcessFilesByMediaDisk(fileKeys,
             new ProcessFilesOnOneMediaDiskHandler(this.ExtractFilesOnOneMediaDisk));
+        this.ExtractFilesInBinaryTable(fileKeys);
     }
 
     private bool IsMergeModule()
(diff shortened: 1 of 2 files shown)


src/DTF/Libraries/WindowsInstaller.Package/InstallPackage.cs
```diff
@@ -191,6 +191,7 @@ public void ExtractFiles(ICollection<string> fileKeys)
     {
         this.ProcessFilesByMediaDisk(fileKeys,
             new ProcessFilesOnOneMediaDiskHandler(this.ExtractFilesOnOneMediaDisk));
+        this.ExtractFilesInBinaryTable(fileKeys);
```
barnson: Can't make this change in v3.x since it's a significant behavior change for all users of this method in DTF. I think the best approach for v3.x is to add a switch to melt to specify the binary output directory. If not specified, don't extract binaries. And that could be all internal to melt rather than a DTF change. Another approach would be to do kinda what dark does and extract files and binaries to their own subdirectories. That's also a behavior change but I could be convinced it's OK...

stunney:
Closing for refactoring based on Bob's comments and discussions via email.

PR #192, opened 2015-01-06
WIXBUG4187 - Melt doesn't export Binary or Icon tables

stunney:
New -xn switch used in conjunction with -x <path> to export files using the Dark style top level directory structure of
File
Binary
Icon

The contents of the File directory will match what melt exported before so there will be a minimal breaking change if people decide to utilize this switch.

### DIFF
--- a/src/DTF/Libraries/WindowsInstaller.Package/InstallPackage.cs
+++ b/src/DTF/Libraries/WindowsInstaller.Package/InstallPackage.cs
@@ -190,7 +190,7 @@ public class InstallPackage : Database
     public void ExtractFiles(ICollection<string> fileKeys)
     {
         this.ProcessFilesByMediaDisk(fileKeys,
-            new ProcessFilesOnOneMediaDiskHandler(this.ExtractFilesOnOneMediaDisk));        
+            new ProcessFilesOnOneMediaDiskHandler(this.ExtractFilesOnOneMediaDisk));
     }
 
     private bool IsMergeModule()
@@ -210,54 +210,16 @@ public class InstallPackage : Database
     /// <returns>A <see cref="M:System.Colletions.Generic.IDictionary`1{string,string}"/> of the filename/fullpath key value pairs of the directory given.</returns>
     public virtual IDictionary<string, string> GetFilePaths(string path, ICollection<string> names = null)
     {
-        IDictionary<string, string> retval = new Dictionary<string, string>(100);
+        IDictionary<string, string> filenameToFullPathMap = new Dictionary<string, string>(100);
 
         foreach(FileInfo fi in new DirectoryInfo(path).GetFiles())
         {
             if (null != names && !names.Contains(fi.Name)) continue;
 
-            retval.Add(fi.Name, fi.FullName);
+            filenameToFullPathMap.Add(fi.Name, fi.FullName);
         }
 
-        return retval;
-    }
-
-    /// <summary>
-    /// Extracts binary data from the `Binary` or `Icon` tables to the designated path.
-    /// </summary>
-    /// <param name="names">The names of the rows to be picked.  If null then all rows will be returned.  If none are matched then none are created.</param>
-    /// <param name="tableName">The name of the table to extract binary data from.  Valid values are "Binary" and "Icon" or a custom table with Name and Data columns of type string and stream respectively.</param>
-    /// <param name="path">The path to extract the files to.  The path must exist before calling this method.</param>
-    [SuppressMessage("Microsoft.Globalization", "CA1308:NormalizeStringsToUppercase")]
-    public virtual void ExtractFilesInBinaryTable(ICollection<string> names, string tableName, string path)
-    {
-        if(!Directory.Exists(path))
-        {
-            throw new ArgumentException(string.Format("The path specified does not exist. {0}", path), "path");
-        }
-
-        View binaryView = this.OpenView("Select `Name`, `Data` FROM `{0}`", tableName);
-        binaryView.Execute();        
-
-        ICollection<string> createdFiles = new List<string>(100);
-
-        for (Record binaryRec = binaryView.Fetch(); binaryRec != null; binaryRec = binaryView.Fetch() )           
-        {
-            string binaryKey = (string)binaryRec[1];
-            Stream binaryData = (Stream)binaryRec[2];
-
-            if (null != names && !names.Contains(binaryKey)) continue; //Skip unspecified values
-
-            createdFiles.Add(binaryKey);
-
-            FileInfo binaryFile = new FileInfo(Path.Combine(path, binaryKey));
-            using (FileStream fs = binaryFile.Create())
-            {
-                for (int a = binaryData.ReadByte(); a != -1; a = binaryData.ReadByte()) fs.WriteByte((byte)a);
-            }
-        }
-
-        ClearReadOnlyAttribute(path, createdFiles);
+        return filenameToFullPathMap;
     }
 
     private void ProcessFilesByMediaDisk(ICollection<string> fileKeys,
@@ -807,7 +769,17 @@ public class InstallPackage : Database
         }
     }
 
-    private void ClearReadOnlyAttribute(string baseDirectory, IEnumerable filePaths)
+    /// <summary>
+    /// Clears the +R flag on the specified files in the directory of your choice.
+    /// </summary>
+    /// <param name="baseDirectory">
+    /// This base path is prepended to each item in the <paramref name="filePaths"/> collection.
+    /// </param>
+    /// <param name="filePaths">
+    /// A <seealso cref="IEnumerable"/> collection of filePaths.  If the objects within override ToString to return a valid path relative to
+    /// <paramref name="baseDirectory"/> this method will work.
+    /// </param>
+    public static void ClearReadOnlyAttribute(string baseDirectory, IEnumerable filePaths)
     {
         foreach(object filePath in filePaths)
         {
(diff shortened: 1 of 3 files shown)


src/DTF/Libraries/WindowsInstaller.Package/InstallPackage.cs
```diff
@@ -190,7 +190,7 @@ public void ExtractFiles()
     public void ExtractFiles(ICollection<string> fileKeys)
     {
         this.ProcessFilesByMediaDisk(fileKeys,
-            new ProcessFilesOnOneMediaDiskHandler(this.ExtractFilesOnOneMediaDisk));
+            new ProcessFilesOnOneMediaDiskHandler(this.ExtractFilesOnOneMediaDisk));
```
barnson: whitespace

src/DTF/Libraries/WindowsInstaller.Package/InstallPackage.cs
```diff
@@ -202,6 +202,64 @@ private bool IsMergeModule()
...
 
+    /// <summary>
+    /// Provides a key (filename) / value (full path) pair where the Binary files can be worked with
+    /// </summary>
+    /// <param name="path">The path to search for files in.  This is NOT recursive.</param>
+    /// <param name="names">The explicit names to search for.  All will be returned if nothing specified.</param>
+    /// <returns>A <see cref="M:System.Colletions.Generic.IDictionary`1{string,string}"/> of the filename/fullpath key value pairs of the directory given.</returns>
+    public virtual IDictionary<string, string> GetFilePaths(string path, ICollection<string> names = null)
+    {
+        IDictionary<string, string> retval = new Dictionary<string, string>(100);
```
barnson: Needs a more meaningful name. "map"?

src/DTF/Libraries/WindowsInstaller.Package/InstallPackage.cs
```diff
@@ -202,6 +202,64 @@ private bool IsMergeModule()
...
+    }
+
+    /// <summary>
+    /// Extracts binary data from the `Binary` or `Icon` tables to the designated path.
+    /// </summary>
+    /// <param name="names">The names of the rows to be picked.  If null then all rows will be returned.  If none are matched then none are created.</param>
+    /// <param name="tableName">The name of the table to extract binary data from.  Valid values are "Binary" and "Icon" or a custom table with Name and Data columns of type string and stream respectively.</param>
+    /// <param name="path">The path to extract the files to.  The path must exist before calling this method.</param>
+    [SuppressMessage("Microsoft.Globalization", "CA1308:NormalizeStringsToUppercase")]
+    public virtual void ExtractFilesInBinaryTable(ICollection<string> names, string tableName, string path)
```
barnson: Also used for Icon table. Maybe "ExtractBinaryFiles"? Actually, I recommend moving this to Melt -- it's useful functionality but right now works only for tables that have binary fields in the second column. DTF should be more general-purpose.

src/DTF/Libraries/WindowsInstaller.Package/InstallPackage.cs
```diff
@@ -202,6 +202,64 @@ private bool IsMergeModule()
...
+            Stream binaryData = (Stream)binaryRec[2];
+
+            if (null != names && !names.Contains(binaryKey)) continue; //Skip unspecified values
+
+            createdFiles.Add(binaryKey);
+
+            FileInfo binaryFile = new FileInfo(Path.Combine(path, binaryKey));
+            using (FileStream fs = binaryFile.Create())
+            {
+                for (int a = binaryData.ReadByte(); a != -1; a = binaryData.ReadByte()) fs.WriteByte((byte)a);
```
barnson: Not terribly concerned with perf in a tool like this but without buffering this could be unpleasant for bigger CA binaries. But all the easy stream copy methods are in .NET v4.<sigh>

src/tools/melt/MeltStrings.resx
```diff
@@ -134,12 +134,16 @@
    -wx[N]     treat all warnings or a specific message ID as an error
               (example: -wx1059 -wx1067)
    -wxall     treat all warnings as errors (deprecated)
-   -x <path>  export binaries from database to this path
+   -x &lt;path&gt;  export binaries from database to this path
               (defaults to output files path)
+   -xn        export contents in much the same manner as dark.exe
```
barnson: Don't want version references if we can avoid. Recommend "export contents of File, Binary, and Icon tables to subdirectories to support patching all types of files."

src/tools/melt/melt.cs
```diff
@@ -252,6 +253,52 @@ private void MeltModule()
         }
 
         /// <summary>
+        /// Extract binary data from tables with a Name and Data column in them.
+        /// </summary>
+        /// <param name="inputPdb">A reference to a <see cref="Pdb"/> as output.  Paths (Data properties) will be modified in this object.</param>
+        /// <param name="package">The installer database to rip from.</param>
+        /// <param name="exportPath">The full path where files will be exported to.</param>
+        /// <param name="tableName">The name of the table to export.</param>
+        private void MeltBinaryTable(ref Pdb inputPdb, InstallPackage package, string exportPath, string tableName)
```
barnson: Only needs to be "ref" if you're replacing the object itself.

src/tools/melt/melt.cs
```diff
@@ -252,6 +253,52 @@ private void MeltModule()
...
+        {
+            if(string.IsNullOrEmpty(tableName))
```
barnson: DTF's different but WiX style is a space between keyword and opening parens. (Not for function invocations, though.)

src/tools/melt/melt.cs
```diff
@@ -34,6 +34,7 @@ namespace Microsoft.Tools.WindowsInstallerXml.Tools
     public sealed class Melt
     {
         private string exportBasePath;
+        private bool exportToWiX4Format;
```
barnson: As above. Maybe "exportToSubdirectories"?

src/tools/melt/melt.cs
```diff
@@ -252,6 +253,52 @@ private void MeltModule()
...
+            }
+            if(null == inputPdb)
+            {
+                throw new ArgumentNullException(@"inputPdb");
+            }
+
+            Table pdbTable = inputPdb.Output.Tables[tableName];
+            if (null == pdbTable)
+            {
+                throw new ArgumentException(string.Format("Table by the name of {0} was not found in the input PDB.", tableName), "tableAndSubDirectoryName");
```
barnson: This will crash with an exception if a table doesn't exist. It doesn't really matter if a particular table doesn't exist, so I suggest just returning if the table's null. (If it mattered, you could make it a bool method and return false.)

src/tools/melt/melt.cs
```diff
@@ -252,6 +253,52 @@ private void MeltModule()
...
+            }
+
+            if (!Directory.Exists(exportPath)) Directory.CreateDirectory(exportPath);
```
barnson: You can skip the Exists check; Directory.CreateDirectory does that for you.

src/tools/melt/melt.cs
```diff
@@ -259,21 +306,40 @@ private void MeltProduct()
...
+            Pdb inputPdb = Pdb.Load(this.inputPdbFile, true, true);
+            
             // extract files from the .msi (unless suppressed) and get the path map of File ids to target paths
             string outputDirectory = this.exportBasePath ?? Environment.GetEnvironmentVariable("WIX_TEMP");
+            string exportBinaryPath = null;
+            string exportIconPath = null;
+
+            if(this.exportToWiX4Format)
+            {
+                exportBinaryPath = Path.Combine(outputDirectory, @"Binary");
```
barnson: Don't need @ strings.

barnson:
Sorry for the nitpicking but it's my job. :smile:

src/DTF/Libraries/WindowsInstaller.Package/InstallPackage.cs
stunney: You are referring to "retval" being changed to "map" or something like that? :)

src/DTF/Libraries/WindowsInstaller.Package/InstallPackage.cs
stunney: I will move to Melt but wouldn't this be handy for custom tables as well in the future that follow the same schema?

src/tools/melt/melt.cs
stunney: Just habbit :)  I'll remove them.

stunney:
I like nitpicking :+1:

src/DTF/Libraries/WindowsInstaller.Package/InstallPackage.cs
stunney: Bah!  VS has been doing this to me a lot lately (I'm sure it is something that I am doing wrong, just don't know WHAT)!

src/DTF/Libraries/WindowsInstaller.Package/InstallPackage.cs
barnson: In DTF, I'd want something more generic, so it didn't require the exact same column definitions. (i.e., pass in the column).

src/DTF/Libraries/WindowsInstaller.Package/InstallPackage.cs
barnson: Yeah, just needs to be 10 percent more meaningful. :)

barnson:
I added a history.md entry for you (and cleaned up some whitespace :) ). Thanks for working through it!

PR #195, opened 2015-01-31
WIXBUG:4589

barnson:
Catch exceptions from Version when passing in arbitrary strings. For bundles, try to recover a three-part version number.

### DIFF
--- a/src/tools/wix/ChainPackageInfo.cs
+++ b/src/tools/wix/ChainPackageInfo.cs
@@ -611,7 +611,23 @@ namespace Microsoft.Tools.WindowsInstallerXml
 
                     if (!Common.IsValidModuleOrBundleVersion(this.Version))
                     {
-                        this.core.OnMessage(WixErrors.InvalidProductVersion(this.PackagePayload.SourceLineNumbers, this.Version, sourcePath));
+                        // not a proper .NET version (i.e., five fields); can we get a valid three-part version number?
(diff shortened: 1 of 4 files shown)


src/tools/wix/ChainPackageInfo.cs
```diff
@@ -611,7 +611,23 @@ private void ResolveMsiPackage(BinderFileManager fileManager, Dictionary<string,
 
                     if (!Common.IsValidModuleOrBundleVersion(this.Version))
                     {
-                        this.core.OnMessage(WixErrors.InvalidProductVersion(this.PackagePayload.SourceLineNumbers, this.Version, sourcePath));
+                        // not a proper .NET version (i.e., five fields); can we get a valid three-part version number?
```
rseanhall: Can we try to get a four-part version number before falling back to three?  Burn does use the fourth part.

barnson: I think we have the technology for that...but the bug says three. Why does the bug say three? We should fire the scribe. I can't think of a reason not to take all four parts except that major upgrades aren't going to work either way...

rseanhall: I don't follow your comment about major upgrades.  Burn will consider 1.0.0.0 to 1.0.0.1 a major upgrade assuming a standard Upgrade table.

PR #208, opened 2015-02-18
proposed a possible solution for Bug #4674

abzaver:
Warning! Check is not performed. This is just the proposed concept. Right now I have found a problem with OverflowException on line 175. And the problem with variable maxPreCabSizeInBytes on line 244. These two problems have to be linked to the 2GB limitation on the size of the cab-file. Here I have no offers... But there may be another problem...

### DIFF
--- a/src/tools/wix/AutoMediaAssigner.cs
+++ b/src/tools/wix/AutoMediaAssigner.cs
@@ -134,6 +134,9 @@ namespace Microsoft.Tools.WindowsInstallerXml
             ulong maxPreCabSizeInBytes;
             int maxPreCabSizeInMB = 0;
             int currentCabIndex = 0;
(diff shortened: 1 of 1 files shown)


abzaver:
Additional studies have shown that the problem is much more difficult than I thought. Changes should affect many modules: Binder.cs, CabinetBuilder.cs, Compiler.cs, CompilerCore.cs. Everywhere on the project can be traced compliance Media = Cabinet. Even in the documentation states that "MaximumUncompressedMediaSize - Size of uncompressed files in each cabinet, in megabytes". The elimination of the problem will require more complex measures.

robmen:
Sounds good. Perhaps have a discussion on wix-devs before undertaking the complete and more complex fix.

PR #212, opened 2015-02-27
Add minimal-verbosity logging

barnson:


robmen:
This is *great*! We should have done this a long time ago.

PR #218, opened 2015-03-04
WIXBUG:4681

barnson:
Corrected return type on the lone WOW64 redirection function that returns a BOOLEAN instead of BOOL.


barnson:
@robmen -- change from http://wixtoolset.org/issues/4681/ would seem to not require an assignment agreement.

robmen:
Yeah, cool. Care to port to wix4?

PR #219, opened 2015-03-06
Add WixStdBaLanguage language and documentation.

barnson:


heaths:
Looks good.

src/ext/BalExtension/wixstdba/WixStandardBootstrapperApplication.cpp
```diff
@@ -13,14 +13,18 @@
...
 static const LPCWSTR WIXSTDBA_VARIABLE_LAUNCH_TARGET_PATH = L"LaunchTarget";
 static const LPCWSTR WIXSTDBA_VARIABLE_LAUNCH_TARGET_ELEVATED_ID = L"LaunchTargetElevatedId";
 static const LPCWSTR WIXSTDBA_VARIABLE_LAUNCH_ARGUMENTS = L"LaunchArguments";
 static const LPCWSTR WIXSTDBA_VARIABLE_LAUNCH_HIDDEN = L"LaunchHidden";
 static const LPCWSTR WIXSTDBA_VARIABLE_LAUNCH_WORK_FOLDER = L"LaunchWorkingFolder";
+
 static const DWORD WIXSTDBA_ACQUIRE_PERCENTAGE = 30;
+
 static const LPCWSTR WIXSTDBA_VARIABLE_BUNDLE_FILE_VERSION = L"WixBundleFileVersion";
+static const LPCWSTR WIXSTDBA_VARIABLE_LANGUAGE = L"WixStdBaLanguage";
```
rseanhall: Looks good, just have some naming comments.  In my wix4 pull request, I created some variables that all started with WixStdBA (capital A).  I'm neutral on which one's better.

There's already `SystemLanguageID` and `UserLanguageID` Burn variables.  This one should probably end in ID, too.

barnson: https://msdn.microsoft.com/en-us/library/vstudio/ms229043%28v=vs.100%29.aspx says you're right.

rseanhall: Awesome, I don't have to update my pull request :)

robmen: Except "Id" is expected instead of "ID". From the same MSDN article:

> Note
>
> The two abbreviations that can be used in identifiers are ID and OK. In Pascal-cased identifiers they should appear as Id, and Ok. If used as the first word in a camel-cased identifier, they should appear as id and ok, respectively.

PR #225, opened 2015-03-29
WIXBUG:4630 - Serialize all variables to the elevated Burn process

rseanhall:


src/burn/engine/variable.cpp
```diff
@@ -889,6 +896,17 @@ extern "C" HRESULT VariableDeserialize(
     __inout SIZE_T* piBuffer
     )
 {
+    return VariableDeserializeEx(pVariables, TRUE, pbBuffer, cbBuffer, piBuffer);
+}
+
+extern "C" HRESULT VariableDeserializeEx(
```
barnson: It's OK to change function signatures in the engine. There's only one instance of VariableDeserialize left.

rseanhall:
Apparently engine function signatures can be changed.

PR #226, opened 2015-03-29
WIXFEAT:3933 - Make WixBundleManufacturer variable writable

rseanhall:


src/burn/engine/registration.cpp
```diff
@@ -616,112 +622,110 @@ extern "C" HRESULT RegistrationSessionBegin(
...
         }
 
         // Publisher: provided by UI
-        if (pRegistration->sczPublisher)
-        {
-            hr = RegWriteString(hkRegistration, REGISTRY_BUNDLE_PUBLISHER, pRegistration->sczPublisher);
-            ExitOnFailure1(hr, "Failed to write %ls value.", REGISTRY_BUNDLE_PUBLISHER);
-        }
+        hr = GetBundleManufacturer(pRegistration, pVariables, &sczPublisher);
+        hr = RegWriteString(hkRegistration, REGISTRY_BUNDLE_PUBLISHER, SUCCEEDED(hr) ? sczPublisher : pRegistration->sczPublisher);
```
barnson: Any reason to skip normal ExitOnFailure handling? If GetBundleManufacturer, it's probably catastrophic anyway...

rseanhall: Just handling it like GetBundleName above.

barnson: If only I had post-commit telepathy to know why that was done that way...

rseanhall: What are the chances that @robmen remembers? http://wix.codeplex.com/SourceControl/changeset/d08bc17758b3abd0fd9c49cfe998dcb02daaf6d7

robmen: Avoid non fatal error, I think.

PR #229, opened 2015-04-09
Add VS2015 properties to the WixVsExtension

heaths:
Resolves WIXBUG:4475

Should also reject PR https://github.com/wixtoolset/wix3/pull/124 in lieu of this PR.

### DIFF
--- a/src/Setup/CommonLib/Folders.wxs
+++ b/src/Setup/CommonLib/Folders.wxs
@@ -208,10 +208,10 @@
     </Fragment>
 
     <Fragment>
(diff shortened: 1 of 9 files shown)


barnson:
Any chance you have VSIX support?

heaths:
@barnson, what do you mean? The VSIX properties are set and they work for VS2015 itself.

barnson:
There's nothing in https://github.com/wixtoolset/wix3/blob/develop/src/ext/VSExtension/wixlib/VsixPackage.wxs. Is it somewhere else?

PR #230, opened 2015-04-09
Create Silent Exec Custom Action

firegiantco:
QuietExec outputs both the command and the standard output to the log, which is problematic if you are passing in a password or similar. The new SilentExec custom action allows one to run an executable without outputting the command or the standard output of the program.

Also, provide a new, more consistent name for quiet exec functionality (WixQuietExec instead of CAQuietExec). It behaves no different than the old name, and the old name still works.

### DIFF
--- a/src/ext/ca/wixca/dll/qtexecca.cpp
+++ b/src/ext/ca/wixca/dll/qtexecca.cpp
@@ -15,8 +15,42 @@
 
 #define OUTPUT_BUFFER 1024
 
+// These old "CA" prefix names are deprecated, and intended to go away in wix 4.0, only staying now for compatibility reasons
+const LPCSTR CAQUIET_TIMEOUT_PROPERTY = "QtExecCmdTimeout";
+const LPCWSTR CAQUIET_TIMEOUT_PROPERTY_WIDE = L"QtExecCmdTimeout";
+const LPCSTR CAQUIET_ARGUMENTS_PROPERTY = "QtExecCmdLine";
+const LPCSTR CAQUIET64_ARGUMENTS_PROPERTY = "QtExec64CmdLine";
+const LPCWSTR CAQUIET_ARGUMENTS_PROPERTY_WIDE = L"QtExecCmdLine";
+const LPCWSTR CAQUIET64_ARGUMENTS_PROPERTY_WIDE = L"QtExec64CmdLine";
+// end deprecated section
+
+// WixCA name quiet commandline argument properties
+const LPCSTR WIX_QUIET_ARGUMENTS_PROPERTY = "WixQuietExecCmdLine";
+const LPCSTR WIX_QUIET64_ARGUMENTS_PROPERTY = "WixQuietExec64CmdLine";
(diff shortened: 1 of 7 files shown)


src/ext/ca/wixca/dll/qtexecca.cpp
```diff
@@ -62,17 +82,17 @@ HRESULT BuildCommandLine(
...
 
     LPWSTR pwzData = NULL;
 
-    if (WcaIsPropertySet("QtExecCmdTimeout"))
+    if (WcaIsUnicodePropertySet(wzPropertyName))
     {
-        hr = WcaGetProperty( L"QtExecCmdTimeout", &pwzData);
-        ExitOnFailure(hr, "failed to get QtExecCmdTimeout");
+        hr = WcaGetProperty(wzPropertyName, &pwzData);
+        ExitOnFailure1(hr, "failed to get %ls", wzPropertyName);
```
heaths: Should use the variadic macros now (just drop the "1" in the macro - they will be removed eventually).

src/ext/ca/wixca/dll/qtexecca.cpp
```diff
@@ -169,6 +178,142 @@ extern "C" UINT __stdcall CAQuietExec64(
...
+{
+    Assert(hInstall);
+    HRESULT hr = S_OK;
+    UINT er = ERROR_SUCCESS;
+
+    hr = WcaInitialize(hInstall, "CAQuietExec64");
+    ExitOnFailure(hr, "failed to initialize");
+
+    hr = ExecCommon64(CAQUIET64_ARGUMENTS_PROPERTY, CAQUIET_TIMEOUT_PROPERTY, TRUE, TRUE);
+    ExitOnFailure(hr, "Failed in  method");
```
heaths: Missing method name.

src/libs/wcautil/qtexec.cpp
```diff
@@ -136,76 +137,79 @@ static HRESULT LogOutput(
...
+            if (bUnicode)
             {
-                ++pEnd;
+                hr = StrAllocConcat(&szLog, (LPCWSTR)pBuffer, 0);
+                ExitOnFailure(hr, "failed to concatenate output strings");
+            }
+            else
+            {
+                hr = StrAllocStringAnsi(&szTemp, (LPCSTR)pBuffer, 0, CP_OEMCP);
+                ExitOnFailure(hr, "failed to allocate output string");
```
heaths: Since you're fixing up the code, may as well capitalize "failed" as new code does.

src/ext/ca/wixca/dll/qtexecca.cpp
Mike-GC: Thanks, fixed.

src/ext/ca/wixca/dll/qtexecca.cpp
Mike-GC: Thanks, fixed.

src/libs/wcautil/qtexec.cpp
Mike-GC: Thanks, fixed.

firegiantco:
Thanks for the comments, Heath. Commits were added to resolve them.

PR #235, opened 2015-04-16
WIXBUG:4734 - Rewrote type-51 CAs using SetProperty.

barnson:


src/chm/documents/customactions/qtexec.html.md
```diff
@@ -54,17 +54,16 @@ When the WixQuietExec (or WixSilentExec) action is run as a deferred custom acti
...
+If you need to set a command line that uses other Windows Installer properties, you must schedule an immediate custom action to set the command line property value and schedule a deferred custom action to run QtExec. The property Id used in the SetProperty custom action must match the Id value used in the deferred custom action. A common use of this pattern for QtExec custom actions is to run an executable that will be installed as a part of the setup. The following is an example of authoring a deferred QtExec custom action that relies on another property value:
 
-    <CustomAction Id="QtExecDeferredExampleWithProperty_Cmd" Property="QtExecDeferredExampleWithProperty"
-                Value="&quot;[#MyExecutable.exe]&quot;" Execute="immediate"/>
+    <SetProperty Id="QtExecDeferredExampleWithProperty" After="CostFinalize"
```
robmen: I think a better option than `CostFinalize` here is to use the name of the CustomAction it is setting the data for. Also, remove the `SetProperty/@Id`.

barnson: 1. OK. 2. I think we need Id given that it's the id of the property and required. :)

PR #238, opened 2015-04-20
Implement ExePackage/CommandLine.

barnson:
- Add WixBundleExecutePackageAction variable: Set to the BOOTSTRAPPER_ACTION_STATE of the package as it's about to executed.
- Add ExePackage/CommandLine to compiler and binder.
- Update Burn to parse CommandLine table in manifest and apply it during ExePackage execution.

### DIFF
--- a/src/burn/engine/exeengine.cpp
+++ b/src/burn/engine/exeengine.cpp
@@ -25,7 +25,7 @@ static HRESULT HandleExitCode(
 
 // function definitions
 
-extern "C" HRESULT ExeEngineParsePackageFromXml(
+static HRESULT ParseExitCodesFromXml(
     __in IXMLDOMNode* pixnExePackage,
     __in BURN_PACKAGE* pPackage
     )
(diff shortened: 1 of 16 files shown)


src/burn/engine/exeengine.cpp
```diff
@@ -25,7 +25,7 @@ static HRESULT HandleExitCode(
 
 // function definitions
 
-extern "C" HRESULT ExeEngineParsePackageFromXml(
+static HRESULT ParseExitCodesFromXml(
```
rseanhall: Static methods don't belong up here.

src/burn/engine/exeengine.cpp
```diff
@@ -165,6 +115,146 @@ extern "C" HRESULT ExeEngineParsePackageFromXml(
...
+    __in BURN_PACKAGE* pPackage
+    )
+{
+    HRESULT hr = S_OK;
+    IXMLDOMNodeList* pixnNodes = NULL;
+    IXMLDOMNode* pixnNode = NULL;
+    DWORD cNodes = 0;
+    LPWSTR scz = NULL;
+
+    // select exit code nodes
```
rseanhall: The WiX style is for comments to be complete sentences, so they should always be capitalized and have periods.

src/burn/engine/exeengine.cpp
```diff
@@ -473,20 +579,61 @@ extern "C" HRESULT ExeEngineExecutePackage(
...
+                break;
+
+            case BOOTSTRAPPER_ACTION_STATE_REPAIR:
+                hr = StrAllocConcat(&sczArguments, commandLineArgument->sczRepairArgument, 0);
+                ExitOnFailure(hr, "Failed to get command-line argument for repair.");
+                break;
+
+            default:
+                hr = E_INVALIDARG;
+                ExitOnFailure1(hr, "Invalid Exe package action: %d.", pExecuteAction->exePackage.action);
```
rseanhall: Changing to ExitOnFailure will make it easier to port to wix4.

src/burn/engine/exeengine.cpp
```diff
@@ -473,20 +579,61 @@ extern "C" HRESULT ExeEngineExecutePackage(
...
+            }
+        }
     }
 
     // build command
-    if (wzArguments && *wzArguments)
+    if (0 < lstrlenW(sczArguments))
```
rseanhall: Why can't we use `if (*sczArguments)`?

src/burn/engine/variable.cpp
```diff
@@ -262,6 +262,7 @@ extern "C" HRESULT VariableInitialize(
         {L"WindowsVolume", InitializeVariableWindowsVolumeFolder, 0},
         {BURN_BUNDLE_ACTION, InitializeVariableNumeric, 0},
         {BURN_BUNDLE_EXECUTE_PACKAGE_CACHE_FOLDER, InitializeVariableString, NULL},
+        {BURN_BUNDLE_EXECUTE_PACKAGE_ACTION, InitializeVariableNumeric, 0},
```
rseanhall: The default should be `NULL`, which requires using `InitializeVariableString`.

src/tools/wix/CommandLineInfo.cs
```diff
@@ -0,0 +1,43 @@
...
+//-------------------------------------------------------------------------------------------------
+
+namespace Microsoft.Tools.WindowsInstallerXml
+{
+    using System;
+
+    /// <summary>
+    /// Utility class for Burn CommandLine information.
+    /// </summary>
+    internal class CommandLineInfo
```
rseanhall: Creating new Info classes makes me sad.

src/tools/wix/Xsd/wix.xsd
```diff
@@ -1232,6 +1233,43 @@
       </xs:attribute>
     </xs:complexType>
   </xs:element>
+  <xs:element name="CommandLine">
+    <xs:annotation>
+      <xs:documentation>Describes additional, conditional command-line arguments for an ExePackage.</xs:documentation>
+      <xs:appinfo>
+        <xse:parent namespace="http://schemas.microsoft.com/wix/2006/wi" ref="ExePackage" />
```
rseanhall: xse:parent elements are only necessary when the parent element is defined in a separate xsd file.

rseanhall: Cool, this would have made things a lot easier.  Can't wait for the MSI property version :)

src/burn/engine/exeengine.cpp
barnson: <sigh> I hate making the compiler's job easier by duplicating declarations.

src/burn/engine/exeengine.cpp
barnson: It might be an empty string.

src/tools/wix/CommandLineInfo.cs
barnson: Mission accomplished. :) Fair point but it turns out there are few *Row classes to model. I thought I had one with WixApprovedExeForElevationRow but then I found ApprovedExeForElevation. :)

src/tools/wix/Xsd/wix.xsd
barnson: We have hundreds of those...oh well. MsiProperty is in the queue (brain queue, not PR queue).

PR #242, opened 2015-04-30
Work around VS2015 RC bug that omits the .wixburn section.

barnson:


barnson:
Yeah, but this seemed safer. Hopefully I can remove both eventually.

PR #243, opened 2015-05-04
WIXBUG:4716 

barnson:
If a .wxl file is missing strings added in v3.10, duplicate the generic string from v3.9. Add support for doing so to locutil.

### DIFF
--- a/src/libs/dutil/locutil.cpp
+++ b/src/libs/dutil/locutil.cpp
@@ -286,6 +286,33 @@ extern "C" HRESULT DAPI LocGetString(
     return hr;
 }
 
+extern "C" HRESULT DAPI LocAddString(
+    __in WIX_LOCALIZATION* pWixLoc,
+    __in_z LPCWSTR wzId,
+    __in_z LPCWSTR wzLocString,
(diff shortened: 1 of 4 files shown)


src/libs/dutil/locutil.cpp
```diff
@@ -286,6 +286,33 @@ extern "C" HRESULT DAPI LocGetString(
...
+    __in BOOL bOverridable
+    )
+{
+    HRESULT hr = S_OK;
+
+    ++pWixLoc->cLocStrings;
+    pWixLoc->rgLocStrings = static_cast<LOC_STRING*>(MemReAlloc(pWixLoc->rgLocStrings, sizeof(LOC_STRING) * pWixLoc->cLocStrings, TRUE));
+    ExitOnNull(pWixLoc->rgLocStrings, hr, E_OUTOFMEMORY, "Failed to reallocate memory for localization strings.");
+
+    LOC_STRING* pLocString = pLocString = pWixLoc->rgLocStrings + (pWixLoc->cLocStrings - 1);
```
rseanhall: `pLocString = pLocString`?  Other than that, looks good.

PR #249, opened 2015-05-15
Do not redownload package payloads when /layout is restarted

heaths:
@barnson reviewed the same change in our internal fork as well.


rseanhall:
Looks good.

PR #250, opened 2015-05-17
WIXFEAT:4382 Added support for "files in use" to WixStdBA

creativbox:


barnson:
Thanks, this is great! Small issue: because it's a behavior change and introduces new loc strings, it needs to be opt-in for WiX v3.x. WixStdBA has a bunch of flags like that; see, for example, https://github.com/wixtoolset/wix3/blob/develop/src/ext/BalExtension/wixext/BalCompiler.cs#L303. Can you add ShowFilesInUse?

creativbox:
I'm still fighting with GIT, but I think I got it right. I've added a "ShowFilesInUse" option the same way as mentioned by you. Please check it out.

src/ext/BalExtension/wixstdba/WixStandardBootstrapperApplication.cpp
```diff
@@ -934,6 +952,26 @@ class CWixStandardBootstrapperApplication : public CBalBaseBootstrapperApplicati
...
+    {
+        if (m_fShowFilesInUse && !m_fPrereq && wzPackageId && *wzPackageId)
+        {
+            //If this is an MSI package, display the files in use page.
+            BAL_INFO_PACKAGE* pPackage = NULL;
+            BalInfoFindPackageById(&m_Bundle.packages, wzPackageId, &pPackage);
+
+            if (pPackage && BAL_INFO_PACKAGE_TYPE_MSI == pPackage->type)
+            {
+                return ShowFilesInUseModal(cFiles, rgwzFiles);
```
rseanhall: Calling it modal is really confusing to me when it's just another page.  I'd really like to see the new page be a modal window instead.

creativbox: It is similar to a modal dialog regarding the other pages and the concept was inspired by dialogs too. That’s where the name was coming from. What would be a better name?

barnson:
I did a bit of style cleanup and added a history entry. It would be great if ThmUtil understood modal pages/dialogs but this is fine, especially in v3.x.

creativbox:
Thanks a lot!

PR #251, opened 2015-05-20
Added ServiceInstall.Account fully qualified documentation

cnadeau:
To help everyone who's looking for answers regarding local accounts and why the service is not properly installing or getting "insufficient privileges"

http://angledontech.blogspot.ca/2011/09/wix-troubleshooting-verify-that-you.html

### DIFF
--- a/src/tools/wix/Xsd/wix.xsd
+++ b/src/tools/wix/Xsd/wix.xsd
@@ -6861,7 +6861,7 @@
       </xs:attribute>
       <xs:attribute name="Account" type="xs:string">
         <xs:annotation>
-          <xs:documentation>The account under which to start the service. Valid only when ServiceType is ownProcess.</xs:documentation>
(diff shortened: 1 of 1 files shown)


src/tools/wix/Xsd/wix.xsd
```diff
@@ -6861,7 +6861,7 @@
       </xs:attribute>
       <xs:attribute name="Account" type="xs:string">
         <xs:annotation>
-          <xs:documentation>The account under which to start the service. Valid only when ServiceType is ownProcess.</xs:documentation>
+          <xs:documentation>The fully qualified account name under which to start the service. This is true even for local accounts (e.g.: ".\LOCAL_ACCOUNT"). Valid only when ServiceType is ownProcess.</xs:documentation>
```
barnson: Instead of "This is true even for local accounts..." how about "Fully qualified names must be used even for local accounts..."?

cnadeau:
You are right, description updated

PR #252, opened 2015-05-22
WIXFEAT:4763 - Add literal flag to Burn variables

rseanhall:


src/burn/engine/variable.h
```diff
@@ -44,6 +44,8 @@ typedef struct _BURN_VARIABLE
     LPWSTR sczName;
     BURN_VARIANT Value;
     BOOL fHidden;
+    // If fLiteral, then when formatting this variable its value should be used as is (don't continue recursively formatting).
```
robmen: Generally, would prefer these to the right of the variable instead above when in a list like this.

rseanhall: I put it above to get Intellisense.  But it looks like putting it to the right also does Intellisense, I'll do that next time.

PR #254, opened 2015-05-23
Add new TransformHistoryEntries task

rseanhall:
to support a merge conflict free way to maintain a top down history file.


robmen:
Well this looks very interesting. @MikeGC would be proud to see his idea come to life. We'll see how well it works.

rseanhall:
Were there any problems with the build?  The only issue I can see is that the new task runs too late - the feed was created from the History.md before it was updated with the file entries.

robmen:
No, build seemed to work just fine. I didn't notice the entries were added too late. Thanks for following up.

PR #255, opened 2015-05-24
Refactor WiX bundle to add a WixStdBA variant, so we always have an e…

barnson:
…asy way of creating a WixStdBA-using bundle when we're mucking about in WixStdBA.

### DIFF
--- a/src/Setup/Bundle/WixStdBundle.wxs
+++ b/src/Setup/Bundle/WixStdBundle.wxs
@@ -1,6 +1,6 @@
 <?xml version='1.0' encoding='utf-8'?>
 <!--
-  <copyright file="Bundle.wxs" company="Outercurve Foundation">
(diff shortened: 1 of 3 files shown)


src/Setup/Bundle/WixStdBundle.wxs
```diff
@@ -0,0 +1,33 @@
...
+
+        <swid:Tag Regid="regid.2008-09.org.wixtoolset" />
+        <Update Location='!(loc.UpdateUrl)' />
+
+        <BootstrapperApplicationRef Id="WixStandardBootstrapperApplication.HyperlinkLicense">
+            <bal:WixStandardBootstrapperApplication LicenseUrl="http://wixtoolset.org/about/license/" />
+        </BootstrapperApplicationRef>
+
+        <Chain>
+            <PackageGroupRef Id='WixPackages' />
```
robmen: Don't we need CLR v4 for WiX toolset?

PR #257, opened 2015-05-26
Always set WixBundleOriginalSourceFolder if not already set

heaths:


barnson:
Also needs to set it if /originalsource is passed on the command line.

heaths:
Seems it would. The two are nearly independent - just that `WixBundleOriginalSourceFolder` uses the parent of `WixBundleOriginalSource` if it was set. If `WixBundleOriginalSourceFolder` is set from the command line, it would override that behavior. This is the same fix we had internally, except that didn't take into account you might only set `WixBundleOriginalSourceFolder` (since our BA was setting both but setting `WixBundleOriginalSourceFolder` was ignored without this fix).

barnson:
I don't understand why they'd be independent. WixBundleOriginalSourceFolder should be WixBundleOriginalSource without the file name. If the engine lets you set WixBundleOriginalSource, it should then set WixBundleOriginalSourceFolder appropriately. Why wouldn't that be the right behavior?

heaths:
That behavior works. By default `WixBundleOriginalSourceFolder` is the parent folder of `WixBundleOriginalSource` unless explicitly set already.

A BA may not set `WixBundleOriginalSource` because, perhaps, the bundle has a different name (like all the different VS editions that share the vast majority of package content). So the BA may only set the `WixBundleOriginalSourceFolder`.

PR #258, opened 2015-05-28
WIXBUG:4774 - fix performance problem when caching

rseanhall:
Resetting the permissions on the cache folder for every payload was expensive. This reduces the time from 30 minutes to 26 seconds.

### DIFF
--- a/history/4774.md
+++ b/history/4774.md
@@ -1,0 +1,1 @@
+* SeanHall: WIXBUG:4774 - fix performance problem when caching thousands of payloads in a single package.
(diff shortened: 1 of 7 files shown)


barnson:
So this speeds it up by only resetting the ACL once per package. Do we need to worry about bundle payloads? Could we do it whenever creating a directory or if it's empty, instead of forcing a new message through the pipe?

rseanhall:
No need to worry about bundle payloads, the code was already resetting the permissions on the directory once (cache.cpp 745).

The code I'm touching is where the directory was being created originally.  So no.

rseanhall:
We did a code review long ago, and the feedback was that the unelevated engine is making security decisions. I'll put this in v4 when I get around to addressing that.

PR #260, opened 2015-05-31
WIXFEAT:4781 - Do a best-effort duplication of strings instead of fai…

barnson:
…ling if there's a missing string.

### DIFF
--- a/src/ext/BalExtension/wixstdba/WixStandardBootstrapperApplication.cpp
+++ b/src/ext/BalExtension/wixstdba/WixStandardBootstrapperApplication.cpp
@@ -1254,29 +1254,30 @@ private: // privates
         hr = LocGetString(m_pWixLoc, L"#(loc.SuccessInstallHeader)", &pLocString);
         if (E_NOTFOUND == hr)
         {
-            hr = LocGetString(m_pWixLoc, L"#(loc.SuccessHeader)", &pLocString);
-            ExitOnFailure(hr, "Failed to load SuccessHeader localization string.");
+            // Duplicate strings, best-effort only.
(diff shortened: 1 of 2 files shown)


src/ext/BalExtension/wixstdba/WixStandardBootstrapperApplication.cpp
```diff
@@ -1254,29 +1254,30 @@ class CWixStandardBootstrapperApplication : public CBalBaseBootstrapperApplicati
...
-
-            hr = LocAddString(m_pWixLoc, L"SuccessInstallHeader", pLocString->wzText, pLocString->bOverridable);
-            ExitOnFailure(hr, "Failed to duplicate localization string for SuccessInstallHeader.");
-
-            hr = LocAddString(m_pWixLoc, L"SuccessRepairHeader", pLocString->wzText, pLocString->bOverridable);
-            ExitOnFailure(hr, "Failed to duplicate localization string for SuccessRepairHeader.");
+            // Duplicate strings, best-effort only.
+            if (SUCCEEDED(LocGetString(m_pWixLoc, L"#(loc.SuccessHeader)", &pLocString)))
+            {
+                hr = LocAddString(m_pWixLoc, L"SuccessInstallHeader", pLocString->wzText, pLocString->bOverridable);
```
rseanhall: What happens when you add a string that already exists?

barnson: It succeeds, though it's essentially hidden. (The original is found first.) So the only failure here is OOM and other about-to-blow-up-anyway errors.

PR #261, opened 2015-06-02
Use consistent metadata for project references to avoid double-building

barnson:


heaths:
Looks fine. I'm surprised that caused a double build.

PR #262, opened 2015-06-08
WiX BA allows multiple clicks of Install button while update check is running

jchoover:
http://wixtoolset.org/issues/4512/

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,3 +1,5 @@
+* JacobHoover: WIXBUG:4512 - fix WiX BA, prevent multiple install clicks
+
 ## WixBuild: Version 3.10.0.1726
 
 * creativbox: WIXFEAT:4382 - Added files-in-use UI to WixStdBA
--- a/src/Setup/WixBA/RootViewModel.cs
+++ b/src/Setup/WixBA/RootViewModel.cs
@@ -37,8 +37,8 @@ namespace Microsoft.Tools.WindowsInstallerXml.UX
         private ICommand closeCommand;
 
         private bool canceled;
-        private InstallationState installstate;
-        private DetectionState detectstate;
+        private InstallationState installState;
+        private DetectionState detectState;
 
         /// <summary>
         /// Creates a new model of the root view.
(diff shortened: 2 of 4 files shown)


heaths:
Looks okay, but wouldn't just disabling the Install button suffice until DetectComplete? That's basically what we do in our BA.

jchoover:
Heath,
  Can't do that cleanly cause we are dog fooding the engine feed updating.  In 3.x I couldn't modify signatures, so we have to call detect a second time if the update feed fails.  I suppose we could make it conditional based off of if Detect succeeded, but it seemed more logical to break this into two different states.

src/Setup/WixBA/RootViewModel.cs
```diff
@@ -110,23 +111,46 @@ public bool Canceled
...
             get
             {
-                return this.state;
+                return this.detectstate;
             }
 
             set
             {
-                if (this.state != value)
+                if (this.detectstate != value)
```
robmen: Should be `detectState`? Two words?

src/Setup/WixBA/RootViewModel.cs
```diff
@@ -110,23 +111,46 @@ public bool Canceled
...
+        }
+
+        /// <summary>
+        /// Gets and sets the installation state of the view's model.
+        /// </summary>
+        public InstallationState InstallState
+        {
+            get
+            {
+                return this.installstate;
```
robmen: Ditto but `installState`.

jchoover: Fixed in updated PR

src/Setup/WixBA/RootViewModel.cs
jchoover: Fixed in updated PR

History.md
```diff
@@ -1,3 +1,5 @@
+* JacobHoover: WIXBUG:4512 - fix WiX BA, prevent multiple install clicks
+
```
barnson: Remember the new history scheme...

PR #263, opened 2015-06-08
Replace default WixUI license with lorem ipsum text.

barnson:


heaths:
Maybe we should update .gitattributes to treat .rtf as text. Not necessary for a merge, but something to think about.

PR #265, opened 2015-06-09
Bug 4532 - InstallPrivileges and InstallScope can specify contradicting values

BMurri:
This is the wix3 version of the fix for this bug. I felt that I couldn't simply return an error when both attributes are supplied, but since the error is opened in 3.x, I felt that I can still return an error.

This code will continue to accept both attributes being present as long as the respective values are compatible. If they are not, an error that recommends keeping only the InstallScope attribute.

The wix4 version of the fix is [here](https://github.com/wixtoolset/wix4/pull/142).

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,3 +1,5 @@
+* BMurri: WIXBUG:4532 - Make it an error for InstallPrivileges and InstallScope to specify contradictory values.
+
 ## WixBuild: Version 3.10.0.1726
 
 * creativbox: WIXFEAT:4382 - Added files-in-use UI to WixStdBA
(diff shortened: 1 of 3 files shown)


History.md
```diff
@@ -1,3 +1,5 @@
+* BMurri: WIXBUG:4532 - Make it an error for InstallPrivileges and InstallScope to specify contradictory values.
```
rseanhall: MikeGC had a great idea of creating a file for each history entry instead of manually updating History.md to avoid merge conflicts.  See https://github.com/wixtoolset/wix3/tree/develop/history for details.

BMurri: That's a great idea. I'll update my pull requests to align with this within 24 hours or so.

PR #266, opened 2015-06-09
WIXBUG:4499 - Bind MBApreq language to correct langid

BMurri:
[WIXBUG:4499](http://wixtoolset.org/issues/4499/)

pull request opened in both wix3 and [wix4](https://github.com/wixtoolset/wix4/pull/143)


barnson:
We can't change behavior in WiX v3.x. Every bundle using WixStdBA goes through this code path. If you wanted to make it an opt-in behavior, we could take it in WiX v3.11 (and v4 of course).

PR #268, opened 2015-06-10
Bug #4792

stunney:
Adding validation of WixPdb to paired MSI to ensure that the two
actually belong together.


robmen:
Two very big things:

1. Only check the PackageCode. That will be an exact match.
2. Don't send messages by throwing exceptions. Write a message like all the other code.

stunney:
1.  Agreed
2.  Crap, my bad.

stunney:
Would you recommend that melt still "abort" upon discovering the mismatch or simply serve up a "WARNING" on the console?

robmen:
For WiX v3.x I think it *has* to be a warning since maybe someone is doing this and it is working for them (magically). It *should* become an error in v4.x.

stunney:
Already added the warning string to the resource file.

stunney:
New PR will be up tonight.

robmen:
You don't need a new PR, you can just update your branch and push back to it. That will update this pull request.

stunney:
updates committed

src/tools/melt/melt.cs
```diff
@@ -308,6 +308,30 @@ private static void MeltBinaryTable(Pdb inputPdb, InstallPackage package, string
             }
         }
 
+        private static bool ValidateMSIMatchesPdb(InstallPackage package, Pdb inputPdb)
+        {
+            const string KEY_PRODUCT_CODE = "ProductCode";
```
heaths: By declaring `const` you're actually causing the string to be allocated multiple times. Either declare as `static readonly` at the class scope or declare here without `const` so the same string instance is used.

stunney: Can't use static readonly at the class level and maintain the usage of the switch statement.  That said, it is a single check so I can easily switch that to an if statement.

stunney:
Updates based on @heaths feedback committed.

PR #273, opened 2015-06-23
Support passing PlatformToolset to the build to force a particular co…

barnson:
…mpiler toolset (e.g., VS2010) regardless of which compiler on the system is the latest.


heaths:
Don't we need to build dutil with the various toolsets to work properly, or is that no longer (or maybe never was) the case?

barnson:
libs_minimal.proj just builds the one arch/toolset needed for the tools tree. libs.proj builds everything.

robmen:
This PR is out of date due to #301 right?

PR #274, opened 2015-06-24
WIXBUG 4791:  (redone) Fix race in lit.exe/light.exe related to wixlibs and la…

jmcooper8654:
…rge binaries.  This pull request has been redone.  The previous one should be ignored.


barnson:
Thanks -- that looks much better. I'll take a look.

jmcooper8654:
I think I know what I did wrong.  I'll try really hard not to repeat that fiasco.

src/tools/wix/BinderFileManager.cs
```diff
@@ -297,7 +300,7 @@ public virtual bool CompareFiles(string targetFile, string updatedFile)
 
                         for (int i = 0; i < targetReadLength; ++i)
                         {
-                            if (targetBuffer[i] != updatedBuffer[i])
+                            if (0 != memcmp(targetBuffer, updatedBuffer, targetReadLength))
```
barnson: This is still inside the loop...? I'd like to see profiling data that it's an improvement before taking in an msvcrt function via P/Invoke.

src/tools/wix/BinderFileManager.cs
```diff
@@ -735,7 +777,60 @@ private static bool CheckFileExists(string path)
...
+
+            while (true)
+            {
+                try
+                {
+                    return func();
+                }
+                catch (IOException)
+                {
+                    var fsw = new FileSystemWatcher(Path.GetDirectoryName(path)) { EnableRaisingEvents = true };
```
barnson: This seems pretty heavyweight. Plus it ends up waiting forever instead of timing out, if there's no event on the file. What if it's a file that's causing the problem but we lock on the directory? Recommend omitting.

History.md
```diff
@@ -1,5 +1,7 @@
 ## WixBuild: Version 3.10.0.1823
 
+* jmcooper8654: WIXBUG:4791 - Fix race in lit.exe/light.exe related to wixlibs and large binaries.
```
barnson: Not needed. The history/*.md entries are added to History.md during the build.

src/tools/wix/BinderFileManager.cs
jmcooper8654: This can be done by a wait-and-retry method too.  Also, a ceiling on the number of tries can be done easily enough.  The FileSystemWatcher just takes the constants out of the equation.

 

I’m certainly open to other techniques to block while another process is writing—this is just one of many.

 

This code is being tested on four servers with four cores, four agents, and 16 GB memory each supporting five teams with twenty-five products—all migrated to MultiProc.  That works out to between 30-50 builds per day.  There hasn’t been a non-code-non-unit test failure in 5 days.

 

It may not be the best, but it does work in our environment.

 

--

John M. Cooper

Leavenworth, KS

src/tools/wix/BinderFileManager.cs
jmcooper8654: Looping over bytes isn’t too bad.  For our payloads, this is significantly faster.  It’s inside the loop, but the blocks are 64k and we’re looping over blocks.

 

Certainly, looping over bytes is just fine.

 

--

John M. Cooper

Leavenworth, KS

rseanhall:

Hi @jmcooper8654, I'm your friendly neighborhood .NET Foundation Pull Request Bot (You can call me DNFBOT). Thanks for your contribution!

TTYL, DNFBOT;

History.md
```diff
@@ -101,4 +101,183 @@ MikeGC: WIXBUG:4878 - fix iniutil memory leak
 
 * BMurri: WIXBUG:3902 - Fix ability to find config files in certain circumstances.
 
+<<<<<<< HEAD
```
rseanhall: There shouldn't be any modifications to History.md.  Also, there are tons of merge commits in here so it would be nice if you rebased onto develop.

PR #278, opened 2015-07-05
WIXBUG:4180 - Add wixtoolset.org EULA service links.

barnson:


Laniusexcubitor:
The links used here lead to a 404

PR #283, opened 2015-07-25
Add support for additional Windows SDKs

heaths:


heaths:
@barnson, any reason this can't go into 3.10? It's only for initialization. VS2015-only machines can't build without these changes (same for wix4 repo).

barnson:
I'll take a look.

rseanhall:
I just built a Win10/VS2015 VM over the weekend, and I can build without this.  Probably had something to do with installing the XP support for C++ option.

jmcooper8654:
For what it is worth, I can build without it too.  The problem I had with XP C++ support turned out to be an incomplete install of VS 2010.  VS 2010 installed without error, but a big chunk of the C/C++ header tree was missing.  I grafted the necessary bits on, applied SP1 and all the updates, and all has been golden.

 

Still, it’s good to be flexible.

 

--

John M. Cooper

Leavenworth, KS

heaths:
I've verified that XP support is checked. That's really just the headers and libs, though (well, mainly libs). On several fresh machines I've had to do this. In fact, over the weekend I'm working on another change to let us build votive using the latest VSSDK since it contains all the previous bits. I really don't want to put VS2010 on this machine just to build WiX (and VSSDK2010 requires VS2010; even if you work around the blocking message, all the binaries VSSDK2010 references are installed by VS2010).

barnson:
I tried it on a Dev14-only VM and OneTimeWixBuildInitialization.proj fails without it. Yet another reason to ship the strong-name signing key...

heaths:
@barnson The private key is only a result of depending on VS2010. I'll have a PR eventually so that you can actually build Votive without VS2010 SDK as well. All the stuff we need should be in all future versions as well.

PR #286, opened 2015-08-12
WIXBUG:4847

jchoover:
- Modified the WixBA UI to display the current status as the title bar is no longer visible to display it.


jchoover:
Created updated PR with the text in the title area.

PR #293, opened 2015-09-01
* RobMen: WIXBUG:4879 - support x64 folders and bind variables in SWID Tags.

firegiantco:


src/ext/TagExtension/wixext/Xsd/tag.xsd
```diff
@@ -118,4 +132,28 @@
       </xs:attribute>
     </xs:complexType>
   </xs:element>
+
```
barnson: Why are these now necessary? @Win64 is a string...

src/ext/TagExtension/wixext/TagBinder.cs
```diff
@@ -198,11 +198,18 @@ private List<WixFileRow> CreateProductTagFiles(Output output)
 
         private static Version CreateFourPartVersion(string versionString)
         {
-            Version version = new Version(versionString);
-            return new Version(version.Major,
-                               -1 < version.Minor ? version.Minor : 0,
-                               -1 < version.Build ? version.Build : 0,
-                               -1 < version.Revision ? version.Revision : 0);
+            try
```
barnson: So a bind-time variable always results in 0.0.0.0?

heaths: Worse: 0.0.-1.-1

robmen: Turns out the tag is generated twice. First time just to get the size about right and second time to actually get the tag fully populated with access to bind variables. Thus the version may start as "0.0" and be updated to bind values. So it works out okay.

PR #301, opened 2015-09-18
Add `DisableVS201x` properties to skip versions of VS during the build.

barnson:
This is useful to save build time during debugging and to diagnose codegen problems in different versions of VS.


BMurri:
Add mention of DefaultToVS to DisableVS.md.

PR #308, opened 2015-10-16
Fix bootstrapper silent hanging behaviour in case access to TMPDIR is denied

Mironenko:
The set of commits fixes the problem, described below:

1. Make TMP directory read-only for user.
2. Run bootstrapper application

Expected: bootstrapper either works normally or informs about an error.
Observed: bootstrapper hangs silently.

### DIFF
--- a/src/burn/engine/cache.cpp
+++ b/src/burn/engine/cache.cpp
@@ -183,6 +183,10 @@ extern "C" HRESULT CacheEnsureWorkingFolder(
     ExitOnFailure(hr, "Failed to calculate working folder to ensure it exists.");
 
     hr = DirEnsureExists(sczWorkingFolder, NULL);
+    if (HRESULT_FROM_WIN32(ERROR_ACCESS_DENIED) == hr)
+    {
+        hr = HRESULT_FROM_WIN32(ERROR_INSTALL_TEMP_UNWRITABLE);
+    }
     ExitOnFailure(hr, "Failed create working folder.");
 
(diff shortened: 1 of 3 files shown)


src/burn/engine/cache.cpp
```diff
@@ -183,6 +183,10 @@ extern "C" HRESULT CacheEnsureWorkingFolder(
     ExitOnFailure(hr, "Failed to calculate working folder to ensure it exists.");
 
     hr = DirEnsureExists(sczWorkingFolder, NULL);
+    if (HRESULT_FROM_WIN32(ERROR_ACCESS_DENIED) == hr)
```
barnson: Anybody have an opinion about adding ERROR_DISK_FULL too?

src/libs/dutil/pathutil.cpp
```diff
@@ -14,6 +14,7 @@
 #include "precomp.h"
 
 #define PATH_GOOD_ENOUGH 64
+static const DWORD TEMP_FILE_CREATE_RETRY_COUNT = 3;
```
robmen: This retry limit is too small given how this function is used. I'd do at least, 300, which would be 30 second wait.

src/burn/engine/cache.cpp
robmen: Seems reasonable.

Mironenko:
Failed to reopen PR after amending so created new PR: https://github.com/wixtoolset/wix3/pull/313

PR #310, opened 2015-10-21
Added description for ERROR_FAIL_NOACTION_REBOOT on Windows XP

PavelAnfimov:
Ticket [here](http://wixtoolset.org/issues/4933/)

Since dutil does not have own string resources (to add them for only this would be a big overhead), let's add at least an English description of the error. That will help many of users to undrestand situation.

### DIFF
--- a/src/libs/dutil/strutil.cpp
+++ b/src/libs/dutil/strutil.cpp
@@ -1169,11 +1169,21 @@ extern "C" HRESULT DAPI StrAllocFromError(
 
     if (0 == cchMessage)
     {
-        ExitWithLastError1(hr, "Failed to format message for error: 0x%x", hrError);
+        if(ERROR_FAIL_NOACTION_REBOOT == HRESULT_CODE(hrError)) // Windows XP
+        {
+            hr = StrAllocString(ppwzMessage, L"No action was taken as a system reboot is required.", 0);
+            ExitOnFailure(hr, "Failed to allocate string for message.");
+        }
+        else
+        {
+            ExitWithLastError1(hr, "Failed to format message for error: 0x%x", hrError);
+        }
     }
-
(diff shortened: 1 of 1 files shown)


PavelAnfimov:
Or let's add to dutil (or burn) own string resources with different translations of the error.

barnson:
As I mentioned in wixtoolset/issues#4933, I think this should go into WixStdBA, not StrUtil. What do you think?

PavelAnfimov:
Are you offer to add in wixstdba.rc string resources (localized) for ERROR_FAIL_NOACTION_REBOOT, and in WixStandardBootstrapperApplication.cpp check if StrAllocFromError returns nothing for ERROR_FAIL_NOACTION_REBOOT and load the description from own resources?

The better solution,i think is add to all standard themes (wxl- files) and load ERROR_FAIL_NOACTION_REBOOT description from corresponding wxl-file for any OS version (via LocLocalizeString())?

barnson:
We don't have localization strings for WixStdBA other than the base English strings, so we need to fall back to OS-provided strings whenever they're available.

PavelAnfimov:
There are no corresponding descriptions for ERROR_FAIL_NOACTION REBOOT on Windows XP. Therefore we need add them manually anywhere in WixStdBA.

1) One way: add localized descriptions to WixStdBA.rc as string resources.
2) Another way: is load localized description from wxl file.

First way is standard for Windows apps.

What do you think about it?

rseanhall:

Hi @PavelAnfimov, I'm your friendly neighborhood .NET Foundation Pull Request Bot (You can call me DNFBOT). Thanks for your contribution!

TTYL, DNFBOT;

barnson:
Implemented in https://github.com/wixtoolset/wix3/pull/413.

PR #311, opened 2015-10-24
Merge remote-tracking branch 'refs/remotes/wixtoolset/develop' into develop

phillHgl:
* PhillHgl: WIXBUG:4932 - Resolved link errors when using the SqlExtension and German localization.  The German translations were already in the Wix Toolset source (in the WixExt_de-de.wxl file) but were not accessible when using the SqlExtension.  This also turned out to be the case for the ComPlusExtension,  FirewallExtension, IIsExtension, MsmqExtension, and UtilExtension.  Each project was updated with its own de-de.wxl file using the strings from WixExt_de-de.wxl.


phillHgl:
This is all messed up, so I am starting over.

PR #317, opened 2015-12-16
Add StrAllocConcatFormatted to concatenate a formatted string to an e…

barnson:
…xisting string.

### DIFF
--- a/history/StrAllocConcatFormatted.md
+++ b/history/StrAllocConcatFormatted.md
@@ -1,0 +1,1 @@
+* MikeGC: Add StrAllocConcatFormatted to concatenate a formatted string to an existing string.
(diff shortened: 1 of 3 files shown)


heaths:
Why was this necessary? You can actually use `StrAllocFormated` for that - I did in several places. The implementation actually supports this by, IIRC, swapping if necessary.

barnson:
@heaths, this is when you have one string and want to concatenate a new formatted string to it. The typical StrUtil reuse doesn't get concatenation.

heaths:
The following works. Use it in a few places.

``` c++
StrAllocString(&wz, L"name: ", 0);
StrAllocFormatted(&wz, L"%ls %hs", wz, "Bob");
```

I supposed it's not intuitive, though. IIRC, can't do that with normal "printf" routines in the CRT.

Code is fine.

barnson:
True. It's a bit voodoo to refer to the variable you're updating...like it's something that works in C# but shouldn't work in C++ because C++ always makes you pay...

heaths:
LGTM

PR #319, opened 2015-12-17
Change Burn's behavior to, instead of skipping all related bundles wh…

barnson:
…en the current bundle is embedded, skip only dependent bundles when the current bundle is a related bundle. (Burn supports embedded mode in cases other than when being executed as a related bundle.)

### DIFF
--- a/history/embedded.related.bundles.md
+++ b/history/embedded.related.bundles.md
@@ -1,0 +1,1 @@
+* BobArnson: Change Burn's behavior to, instead of skipping all related bundles when the current bundle is embedded, skip only dependent bundles when the current bundle is a related bundle. (Burn supports embedded mode in cases other than when being executed as a related bundle.)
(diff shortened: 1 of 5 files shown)


heaths:
We put this code in there originally to prevent some issue. While I don't see any specific problems, I'd like a couple others to look at this.

barnson:
We believe the original code used embedded mode to detect when running as a related bundle. The fix explicitly detects running as a related bundle instead of using embedded mode as a proxy.

heaths:
We had some issues with slipstreaming bundles, where we embed a bundle (e.g. VSUpdate) in the parent bundle to make sure it's registered for other bundles that relate to it. The change was made long enough ago, though, I don't remember all the details.

barnson:
Rob and I spent lots of time walking through the scenarios and what the original change did. This should be behavior-neutral for related bundles and lets a bundle (or app) using embedded mode retain the ability to trigger related bundles.

PR #320, opened 2015-12-21
Make CBalBaseBootstrapperApplication::PromptCancel usefully overrideable

barnson:


src/libs/balutil/inc/BalBaseBootstrapperApplication.h
```diff
@@ -684,14 +684,16 @@ class CBalBaseBootstrapperApplication : public IBootstrapperApplication
         ReleaseNullObject(m_pEngine);
     }
 
+protected:
+    CRITICAL_SECTION m_csCanceled;
```
heaths: This feels dirty exposing fields like that. Seems you could return the CS like so:

``` C++
protected:
  CRITICAL_SECTION& GetCriticalSection() const
  {
    return m_csCanceled;
  }
```

The BOOL I'm less concerned about since it can only be `FALSE` (0) or `!FALSE`. But a CS could be reinitialized. Then again, the BA would only be hurting itself but seems we should follow good coding practices.

PR #325, opened 2016-01-27
Fix for /layout after clean room implementation.

jchoover:
This allows my limited testing to put the layout in the correct location.


src/burn/engine/apply.cpp
```diff
@@ -950,11 +952,15 @@ static HRESULT LayoutBundle(
 {
     HRESULT hr = S_OK;
     LPWSTR sczBundlePath = NULL;
+    LPWSTR sczBundleSourcePath = NULL;
```
jchoover: /sigh, I neglected to free this.

PR #332, opened 2016-02-15
Fix up bad/old links in Learning WiX topic.

barnson:


barnson:
Fixes wixtoolset/issues#4397.

PR #333, opened 2016-02-22
Fix #5217: Ignore failure to check domain existence on ConfigureUsers

nirbar:
Reset er and hr variables when deciding to continue after failing to check existence of the domain

### DIFF
--- a/src/ext/ca/serverca/scasched/scauser.cpp
+++ b/src/ext/ca/serverca/scasched/scauser.cpp
@@ -540,6 +540,8 @@ HRESULT ScaUserExecute(
             ueUserExists = USER_EXISTS_INDETERMINATE;
             hr = HRESULT_FROM_WIN32(er);
             WcaLog(LOGMSG_VERBOSE, "Failed to check existence of domain: %ls, user: %ls (error code 0x%x) - continuing", wzDomain, psu->wzName, hr);
+			hr = S_OK;
+			er = ERROR_SUCCESS;
         }
 
         if (WcaIsInstalling(psu->isInstalled, psu->isAction))
--- a/src/ext/ca/serverca/scasched/scauser.cpp
+++ b/src/ext/ca/serverca/scasched/scauser.cpp
@@ -515,7 +515,7 @@ HRESULT ScaUserExecute(
         if (wzDomain && *wzDomain)
         {
             er = ::DsGetDcNameW(NULL, wzDomain, NULL, NULL, NULL, &pDomainControllerInfo);
-            if (HRESULT_FROM_WIN32(er) == RPC_S_SERVER_UNAVAILABLE)
+            if (RPC_S_SERVER_UNAVAILABLE == er)
             {
                 // MSDN says, if we get the above error code, try again with the "DS_FORCE_REDISCOVERY" flag
                 er = ::DsGetDcNameW(NULL, wzDomain, NULL, NULL, DS_FORCE_REDISCOVERY, &pDomainControllerInfo);


src/ext/ca/serverca/scasched/scauser.cpp
```diff
@@ -515,7 +515,7 @@ HRESULT ScaUserExecute(
         if (wzDomain && *wzDomain)
         {
             er = ::DsGetDcNameW(NULL, wzDomain, NULL, NULL, NULL, &pDomainControllerInfo);
-            if (HRESULT_FROM_WIN32(er) == RPC_S_SERVER_UNAVAILABLE)
+            if (RPC_S_SERVER_UNAVAILABLE == er)
```
robmen: MSDN documentation doesn't say anything about this error code. While I agree it looks like it was wrong before, what is this change going to make work now? The comment on line 520 confuses this issue.

nirbar: Agree that this specific error code may not be the only one relevant to repeat the call with DS_FORCE_REDISCOVERY flag. I suggest doing it for any error code except ERROR_INVALID_DOMAINNAME, ERROR_NOT_ENOUGH_MEMORY

src/ext/ca/serverca/scasched/scauser.cpp
```diff
@@ -540,6 +540,8 @@ HRESULT ScaUserExecute(
             ueUserExists = USER_EXISTS_INDETERMINATE;
             hr = HRESULT_FROM_WIN32(er);
             WcaLog(LOGMSG_VERBOSE, "Failed to check existence of domain: %ls, user: %ls (error code 0x%x) - continuing", wzDomain, psu->wzName, hr);
+			hr = S_OK;
+			er = ERROR_SUCCESS;
```
robmen: Tabs instead of whitespaces?

nirbar: No, that's just a Notepad++ thing
The change is resetting er and hr to success values. 
Otherwise the error code is dragged on to the return when the function has nothing more to do (subsequent 'if' statements are all false)

src/ext/ca/serverca/scasched/scauser.cpp
robmen: I'm not excited about turning on this code path for the first time ever (since it seems the existing error check should never be true) without a good idea when it should be called. Have you tested your suggestion and found it works correctly?

nirbar: No, I edited that area because er and hr dragged on their error values to the return value.
However, MSDN does state that retrying the function with DS_FORCE_REDISCOVERY is the thing to do when the cached domain is not available.

robmen: Yes. So does that mean only retry on the error `ERROR_NO_SUCH_DOMAIN`? The others suggest API failure not transient caching issues... right?

src/ext/ca/serverca/scasched/scauser.cpp
robmen: Right. I'm commenting that the whitespace appears to be incorrect. Were tabs used instead of spaces or is GitHub just rendering it funny?

nirbar: It's tabs.

src/ext/ca/serverca/scasched/scauser.cpp
nirbar: At least ERROR_NO_SUCH_DOMAIN and RPC_S_SERVER_UNAVAILABLE. 
MSDN doesn't explictly state all possible error codes

heaths:
Did you actually experience a problem with this code though? Bob wrote this internally a long time ago and it's been working fine in ask the environments where VS is installed. We've never had a problem with this code since.

nirbar:
I actually had issues with it when removing on a machine where internet is
disconnected (though intranet isn't):

*Action start 8:11:27: ConfigureUsers.*
*ConfigureUsers:  Error 0x800706ba: failed to add/remove User actions*
*CustomAction ConfigureUsers returned actual error code 1603 (note this may
not be 100% accurate if translation happened inside sandbox)*

The annoying part is, that ConfigureUsers isn't supposed to be doing
anything during removal of this product...


*Nir Bar*
*Panel::Software Architecture and Development*
*Mobile: +972-54-2490321*
*Mail: nir.bar@panel-sw.co.il <nir.bar@panel-sw.co.il>*
*Web**: www.panel-sw.co.il <http://www.panel-sw.co.il>*

*   - C++ On Windows, Linux and Embedded Platforms*
*   - WiX & InstallShield*

* - Build & Continuous Integration*

2016-04-08 18:46 GMT+03:00 Heath Stewart <notifications@github.com>:

rseanhall:

Hi @nirbar, I'm your friendly neighborhood .NET Foundation Pull Request Bot (You can call me DNFBOT). Thanks for your contribution!

TTYL, DNFBOT;

PR #336, opened 2016-02-27
Optionally allow duplicate directory identifiers

heaths:
Resolves wixtoolset/issues#5227 to allow merging wixlibs from swix into
products linked using traditional wix.


robmen:
:-1: I left a comment on the originating issue that I don't think this hack should exist in WiX v3.x. The real feature exists in WiX v4.

heaths:
Deprecated switches per discussion with @barnson. I ended up using the same format in the usage text as `-bf`.

PR #339, opened 2016-03-06
WIXBUG:5234 - WiX 3.10.1 failing to update to 3.10.2

rseanhall:
Addresses wixtoolset/issues#5234.

Make Burn grab a file handle to the original bundle so it can still access compressed payloads even if the original bundle was deleted after initialization (e.g. when a bundle uses Burn's built-in update mechanism).

It's a little more complicated than I like, since Burn has to acquire attached containers that aren't actually attached and it only supported acquiring by copying or downloading a file.


jmcooper8654:
Looking at it, it should work.  I'll create another local fork and put it to the test Monday morning.

PR #341, opened 2016-03-22
WIXBUG:5238 - Get the engine's file handle as soon as possible

rseanhall:
Short term fix for wixtoolset/issues#5238.

PR #343, opened 2016-04-04
Use inherited file handles to remove race conditions

rseanhall:
Related to wixtoolset/issues#5238 and wixtoolset/issues#5234.

Includes the commit from #341.


barnson:
:100: Internet Points for coming up with a change that crosses Burn and ThmViewer without including ThmUtil via WixStdBA. :)

PR #344, opened 2016-04-07
Add internal WixBundleUILevel variable

jozefizso:
Internal `WixBundleUILevel` variable is set to the value of `BOOTSTRAPPER_DISPLAY` enum.

Burn engine will set `WixBundleUILevel` value based on `pEngineState->command.display` field after parsing command line options. Parameters `/quiet` and `/passive` can change UI level value.

Fixes https://github.com/wixtoolset/issues/issues/3789


barnson:
@jozefizso can you sign the .NET Foundation CLA? http://wixtoolset.org/development/assignment-agreement/

PR #346, opened 2016-04-18
Prevent bad use of ".." in payload names. Fixes wixtoolset/issues#4265.

barnson:


robmen:
We chose not to take this for WiX v3. The fix will be in next WiX v4 build.

PR #348, opened 2016-04-20
Change a button label in Estonian to match description

Telgat:
The Estonian text used in exit and error descriptions was not the same as the actual text used on the finish button (The description said to push the "Lõpeta" button while the button was actually labeled "Valmis")
Translated to English it would be like "Done" vs "Finish"


barnson:
Thanks!

PR #349, opened 2016-04-27
burn engine: Always use the bundle source path for all purposes.

res2k:
When layouting we need the original source path in all cases: for
copying the bundle to the layout directory as well as for checking
whether we're layouting to the bundle location.

This fixes issue #5277.

### DIFF
--- a/history/5277.md
+++ b/history/5277.md
@@ -1,0 +1,1 @@
+* FRichter: WIXBUG:5277 - burn engine: Always use the bundle source path for all purposes. The original source path is needed in all cases: for copying the bundle to the layout directory as well as for checking whether we're layouting to the bundle location.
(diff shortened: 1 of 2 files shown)


rseanhall:
This looks good.  We need to complete the switch to .NET Foundation before we merge it.

Long term, it would be nice to use the file handle when copying instead of grabbing a new handle.  Also, we normally add a history file for each bug fix.

barnson:
wixtoolset/issues#5277

barnson:
@res2k now that we've moved to .NET Foundation, can you sign the .NET Foundation CLA. I promise this is the last time you'll need to go through a process... and the .NET Foundation one is fully automated so it should go quickly.

barnson:
http://wixtoolset.org/development/assignment-agreement/

rseanhall:
Thanks, this has been merged into 3.10.3 and 3.11.

PR #351, opened 2016-05-03
Workaround GDI+ security vulnerability

robmen:
GDI+ fails in some scenarios when ::SetDefaultDllDirectories() is used
to protect a process from DLL hijacking. All WinForms based BAs hit this
problem. To workaround the failure, all Burn processes that load BAs
will no longer protect themselves from DLL hijacking. In particular, the
clean room and package cache processes will be unsecure. Fortunately,
the design of those processes should be safe without the extra
protection.

Fixes wixtoolset/issues#5282

### DIFF
--- a/src/burn/engine/engine.cpp
+++ b/src/burn/engine/engine.cpp
@@ -54,6 +54,29 @@ static HRESULT Restart();
 
 // function definitions
 
+extern "C" BOOL EngineInCleanRoom(
+    __in_z_opt LPCWSTR wzCommandLine
+    )
+{
+    // Be very careful with the functions you call from here.
+    // This function will be called before ::SetDefaultDllDirectories()
+    // has been called so dependencies outside of kernel32.dll are
+    // very likely to introduce DLL hijacking opportunities.
+
+    static DWORD cchCleanRoomSwitch = lstrlenW(BURN_COMMANDLINE_SWITCH_CLEAN_ROOM);
+
+    // This check is wholly dependent on the clean room command line switch being
+    // present at the beginning of the command line. Since Burn is the only thing
+    // that should be setting this command line option, that is in our control.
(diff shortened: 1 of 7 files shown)


src/burn/engine/engine.cpp
```diff
@@ -65,6 +65,79 @@ static HRESULT Restart();
...
+
+    if (!::StringFromGUID2(guidBundleId, wzBundleId, countof(wzBundleId)))
+    {
+        hr = E_OUTOFMEMORY;
+        ExitOnRootFailure(hr, "Failed to convert bundle guid into string.");
+    }
+
+    for (DWORD i = 0; i < 2; i++)
+    {
+        hr = CacheGetCompletedPath(fTryPerMachine, wzBundleId, &sczCacheFolder);
```
rseanhall: Calling into the engine code before `AppInitialize` makes me very uncomfortable.  Could a bundle running in the package cache spawn off a clean room process from the package cache (doesn't copy itself anywhere, just runs itself)?

robmen: That's a pretty good idea. Let me look into that.

rseanhall:
Launching the extra process was totally worth it.  :+1:

robmen:
Agreed, @rseanhall. Brilliant suggestion. Thank you.

src/libs/dutil/apputil.cpp
```diff
@@ -81,6 +81,12 @@ extern "C" void DAPI AppInitialize(
     }
 }
 
+extern "C" void DAPI AppInitializeUnsafe()
+{
+    ::HeapSetInformation(NULL, HeapEnableTerminationOnCorruption, NULL, 0);
+
```
barnson: whitespace

robmen: Fixed in latest amend.

PR #352, opened 2016-05-05
Enable the WiX native SDK when the new Visual C++ Build Tools bundle …

barnson:
…is installed.

### DIFF
--- a/history/5279.md
+++ b/history/5279.md
@@ -1,1 +1,1 @@
-* @barnson: Enable the WiX native SDK when the new Visual C++ Build Tools bundle is installed. Fixes wixtoolset/issues#5279.
+* @barnson: Enable the WiX native SDK when the new Visual C++ Build Tools bundle is installed. Fixes WIXBUG:5279.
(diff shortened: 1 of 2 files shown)


history/5279.md
```diff
@@ -0,0 +1 @@
+* @barnson: Enable the WiX native SDK when the new Visual C++ Build Tools bundle is installed. Fixes wixtoolset/issues#5279.
```
robmen: I think it'd be best if we keep using the WIXBUG:### nomenclature.

robmen:
Should add comment in PR that says which wixtoolset/issue it fixes so GitHub automatically handles the closing.

history/5279.md
barnson: Don't we want the auto association? Add `wixtoolset/issues#5279` in commit comment or PR comment instead?

rseanhall: I'm with Rob here.  I'd like to keep the `committer: WIXBUG:###` nomenclature.  And PR comment instead of commit comment so we can have a link to the issue from the PR.

barnson: Commit comment works. See aae66628e918a5e2c98232134225fb01266370de. (And it's a good thing because it means we can do it offline.)

PR #355, opened 2016-05-15
Create references to files referenced in EventManifest.

barnson:
Fixes wixtoolset/issues#5295.


barnson:
Turns out the attributes are formatted to allow non-File/@Id references (e.g., file searches).

PR #357, opened 2016-05-18
Add ineritable attribute to PermissionEx to control inheritance flags

lordjeb:
Fixes wixtoolset/issues#2006
Fixes wixtoolset/issues#2751
Fixes wixtoolset/issues#2580


src/ext/UtilExtension/wixext/Data/tables.xml
```diff
@@ -134,6 +134,8 @@
                 category="text" description="Domain half of user account to secure"/>
         <columnDefinition name="User" type="string" length="255" primaryKey="yes" modularize="property"
                 category="text" description="Username half of user account to secure"/>
+        <columnDefinition name="Inheritable" type="number" length="1" modularize="column"
```
robmen: Unfortunately, cannot add columns to a table. This will break patching. It appears this table is old and doesn't have the generic Attributes column we stared adding to somewhat future proof ourselves for these sorts of changes.

I'm afraid this change will need to move to WiX v4 only.

src/ext/ca/wixca/dll/secureobj.cpp
```diff
@@ -586,6 +591,7 @@ extern "C" UINT __stdcall ExecSecureObjects(
     DWORD dwRevision = 0;
     LPWSTR pwzUser = NULL;
     DWORD dwPermissions = 0;
+    int isInheritable = 0;
```
robmen: This is really a BOOL right?  If so, then you'd want `BOOL fInheritable` and if it really should be a numeric, then you'd want a DWORD so `DWORD dwInheritable`.

Actually, based on comment above, probably best to make this column the `Attributes` column and make Inheritable a bit flag in that column. Future proof ourselves, finally.

history/2006.md
```diff
@@ -0,0 +1 @@
+* LordJeb: WIXFEAT:2006 - Add ineritable attribute to PermissionEx to control inheritance flags
```
robmen: Typo: `ineritable`

robmen:
So, this is cool but ultimately I think the break change means it needs to move to WiX v4 and should add it as a bit flag in an `Attributes` column to be similar to other tables.

lordjeb:
@robmen What about a simple fix to apply inheritance automatically for registrykey (in addition to createfolder). I realize that is not a backward compatible change, but there's a pretty strong case to be made that the behavior is a bug.

Either way, agree with your suggestions, and I will make them to the wix4 version of these changes.

robmen:
"not a backward compatible change" is a non-starter in WiX v3.x unless we're fixing a security vulnerability or a crash (which often means security vulnerability as well).

PR #359, opened 2016-05-20
Fix bug where the file handles were no longer put on the command line

rseanhall:
While debugging this, I found another issue - apparently the new way of launching the process with the file handles causes the Executable path to be in wzCommandLine so there's an extra argument now.  We'll have to fix that, too.

### DIFF
--- a/src/burn/engine/engine.cpp
+++ b/src/burn/engine/engine.cpp
@@ -393,6 +393,7 @@ static HRESULT RunUntrusted(
     LPWSTR wzCleanRoomBundlePath = NULL;
     LPWSTR sczCachedCleanRoomBundlePath = NULL;
     LPWSTR sczParameters = NULL;
+    LPWSTR sczFullParameters = NULL;
     LPWSTR sczFullCommandLine = NULL;
     STARTUPINFOW si = { };
     PROCESS_INFORMATION pi = { };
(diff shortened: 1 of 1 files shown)


src/burn/engine/engine.cpp
```diff
@@ -435,21 +439,21 @@ static HRESULT RunUntrusted(
 
     // The clean room switch must always end up at the front of the command line so
     // the EngineInCleanRoom function will operate correctly.
-    hr = StrAllocFormattedSecure(&sczParameters, L"-%ls=\"%ls\" %ls", BURN_COMMANDLINE_SWITCH_CLEAN_ROOM, sczCurrentProcessPath, wzCommandLine);
-    ExitOnFailure(hr, "Failed to allocate parameters for unelevated process.");
+    hr = StrAllocFormattedSecure(&sczFullParameters, L"-%ls=\"%ls\" %ls", BURN_COMMANDLINE_SWITCH_CLEAN_ROOM, sczCurrentProcessPath, sczParameters);
+    ExitOnFailure(hr, "Failed to append %ls", BURN_COMMANDLINE_SWITCH_CLEAN_ROOM);
```
robmen: `prepend` not `append`, right?

robmen:
Do you want me to take this (with the logging fix) or should we wait for the extra argument fix as well?

src/burn/engine/engine.cpp
rseanhall: Yeah, all three of these should be prepend.

rseanhall:
I'll file a bug for both of these and resubmit.

PR #360, opened 2016-05-21
Also checking exit codes as HResult (#5299)

Laniusexcubitor:
I'm doing the check also as HResult (exactly as it is done for the default values some lines afterwards).

### DIFF
--- a/src/burn/engine/exeengine.cpp
+++ b/src/burn/engine/exeengine.cpp
@@ -773,7 +773,8 @@ static HRESULT HandleExitCode(
         {
             typeCode = pExitCode->type;
         }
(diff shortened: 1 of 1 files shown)


robmen:
This change introduces new side effects that would break backwards compatibility with WiX v3. If you really need to support both the HRESULT and non-HRESULT values, just enter them both as values in the `ExitCode` element.

Laniusexcubitor:
Ok, I did not try to add the HResult in 3.10. Just in 3.8. With some research I found out this was fixed in 3.9. 
Should have also tested this obvious workaround with 3.10. Sorry…
Will try to get the update to 3.10 approved asap.

PR #364, opened 2016-06-02
Allow custom ProjectDefineConstants defined in properties

kzu:
Fixes issue https://github.com/wixtoolset/issues/issues/5313


kzu:
Can use $(DefineConstants). Closing!

PR #365, opened 2016-06-07
Remove copyleft header from templates

robmen:
Templates are meant to be starting points for developers to build on.
They should not be copyleft or that would force any project using our
templates to release their source code.


barnson:
Cherry pick to v3.10.3 branch?

robmen:
Yep. With @rseanhall's latest fix as well.

PR #368, opened 2016-07-10
Wixbug5334

jchoover:


jchoover:
Hmm, this PR isn't clean.  Only the last commit should be needed.

PR #369, opened 2016-07-11
Fixed WIXBUG5334 where the provider key was being returned instead of BundleId

jchoover:
…the bundle id.

### DIFF
--- a/src/libs/dutil/butil.cpp
+++ b/src/libs/dutil/butil.cpp
@@ -217,7 +217,7 @@ HRESULT DAPI BundleEnumRelatedBundle(
                 hr = ::StringCchLengthW(sczUninstallSubKey, STRSAFE_MAX_CCH, reinterpret_cast<UINT_PTR*>(&cchUninstallSubKey));
                 ExitOnFailure(hr, "Failed to calculate length of string");
 
-                hr = ::StringCchCatNExW(lpBundleIdBuf, MAX_GUID_CHARS + 1, sczUninstallSubKey, cchUninstallSubKey, NULL, NULL, STRSAFE_FILL_BEHIND_NULL);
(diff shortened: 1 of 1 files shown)


src/libs/dutil/butil.cpp
```diff
@@ -214,15 +212,12 @@ HRESULT DAPI BundleEnumRelatedBundle(
...
-            hr = ::StringCchLengthW(szProviderKey, STRSAFE_MAX_CCH, reinterpret_cast<UINT_PTR*>(&cchProviderKey));
-            ExitOnFailure(hr, "Failed to calculate length of string");
-
             if (lpBundleIdBuf)
             {
-                hr = ::StringCchCatNExW(lpBundleIdBuf, MAX_GUID_CHARS + 1, szProviderKey, cchProviderKey, NULL, NULL, STRSAFE_FILL_BEHIND_NULL);
+                hr = ::StringCchLengthW(sczUninstallSubKey, STRSAFE_MAX_CCH, reinterpret_cast<UINT_PTR*>(&cchUninstallSubKey));
+                ExitOnFailure(hr, "Failed to calculate length of string");
+
+                hr = ::StringCchCatNExW(lpBundleIdBuf, MAX_GUID_CHARS + 1, sczUninstallSubKey, cchUninstallSubKey, NULL, NULL, STRSAFE_FILL_BEHIND_NULL);
```
barnson: StringCchCatNExW is going to fail after the first GUID if the limit is MAX_GUID_CHARS. Any reason not to use StringCchCopy*?

jchoover: Brain fart, and it's what I had previously done incorrectly.

Will update accordingly.

jchoover: Updated.

PR #375, opened 2016-08-24
Use correct NuGet API endpoint

jozefizso:
First time project initialization may fail on `nuget.exe install` when user has the NuGet 3 installed and the global config file uses v3 API endpoint because the WiX project bundle NuGet 2 in tools.

This adds the v2 API endpoint to NuGet configuration file.

### DIFF
--- a/nuget.config
+++ b/nuget.config
@@ -6,4 +6,8 @@
   <config>
     <add key="repositoryPath" value="packages" />
   </config>
+  <packageSources>
+    <clear />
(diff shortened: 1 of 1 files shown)


rseanhall:
I would prefer that we update the checked in nuget.exe like we did in wix4 (https://github.com/wixtoolset/wix4/pull/210).  I guess it wouldn't hurt to also update the config like you have done here to point to the v3 api.

PR #377, opened 2016-08-27
Update NuGet client to v3.4.4

jozefizso:
Use the latest NuGet client.


rseanhall:
Thanks!

jmcooper8654:
Note, when building packages, 3.4.4 has a bug.  They’re using a new XML parser, and it doesn’t like any XML entities in the nuspec file.  But performance for downloading packages is much improved.

--

John M. Cooper

PR #378, opened 2016-09-25
Fix entire file being bold

Daniel15:
The final `\` in `c:\Program Files` appeared to escape the `<` in `</strong>`, resulting in the entire page being bold. I just added another slash to escape it.

Moved from https://github.com/wixtoolset/web/pull/19


rseanhall:
Thanks!

PR #381, opened 2016-10-19
Adding Albanian translation

ujdhesa:
Hello,

I’ve translated wix3 in Albanian. My only concern is about comments within the file, as I’ve seen other translations not having them. 

TIA,
Besnik


ujdhesa:
Forgot to change the locale value from en-US to sq-SQ. Apologies.

barnson:
Please add an entry in the history directory and add the .wxl file to UIExtension.wixproj so it's built into WixUIExtension.dll.

robmen:
Is `sq-SQ` really the right locale? Should it be `sq-AL`? I don't find reference to `sq-SQ` anywhere.

ujdhesa:
@robmen, SQ is short for Shqipëri (country name, as called from natives); AL is short for Albania (country name, as called from outsiders). The locale shoud be defined by al_AL or sq_SQ… I tend to see sq_AL kind of inconsistent…
If possible, I would use plain sq. If the ab_AB format is mandatory for wixtoolset, than either sq_SQ or sq_AL are fine, but I guess here we'd have to use the one adopted from Windows. Having no experience with the given platform, I'd need some guiding about that.

ujdhesa:
A couple of searches reveal that Windows uses sq_AL, so I'm changing the label to that and proceed accordingly.

@barnson, I apologise for being late with completing your request, but I got confused with history directories.

robmen:
Thanks, need the same updates to wixtoolset/wix4#222 as well, please.

ujdhesa:
@robmen, I’ll do the same for  wix4. 

Could you point me to an example of history entry requested? I see https://github.com/ujdhesa/wix3/blob/develop/history/utilextensiontranslations.md but I'm confused about where to save my new entry. Should I just update https://github.com/ujdhesa/wix3/blob/develop/history/utilextensiontranslations.md with an entry for sq, or should I save it under a new file (of what type)?

robmen:
Ahh, yeah, hard to tell what to add when all the examples have already been processed. Add a file with the `.md` extension to the `history\` folder. The filename doesn't matter, just needs to be unique. Most use the PR number.

Then add a single line (no leading spaces, trailing whitespace ignored) like so:

    * ujdhesa - Add Albanian translation.

During the build, your work will be incorporated into the `History.md` and flow from there.

ujdhesa:
Many thanks!

ujdhesa:
@robmen, my last commit is 24 commits behind wixtoolset:develop . Do I need to take any (git) action related to that, in order to have my PR processed?

robmen:
No. I'll rebase it for you if you are not comfortable doing so yourself.

ujdhesa:
Please do so. Thank you!

PR #384, opened 2016-12-11
Remove MPF project code we're not using and likely won't, in v3.

barnson:


barnson:
"Fixes" https://github.com/wixtoolset/issues/issues/5398

PR #385, opened 2016-12-11
Provide better diagnosis for no-ops.

barnson:
Add logging to WixRemoveFoldersEx for the one early-exit case.


barnson:
Might explain https://github.com/wixtoolset/issues/issues/5389.

PR #386, opened 2016-12-17
Performance Counters - multi-instance & average-base - fixes bug##5457

sjswoboda:
Instead of creating a performance counter object, retrieves the existing one. This allows support for multi-instance counters, as multi-instance counters cannot be created without an instance name.


sjswoboda:
Closed because if there is an instance, then GetCounters() does not work; if there is no instance, then GetCounters() must be called -- regardless of what kind of performance counter category it is.

PR #389, opened 2016-12-21
Document that some properties are available in WiX v3.11 and later.

barnson:


barnson:
https://github.com/wixtoolset/issues/issues/5461

PR #391, opened 2017-01-20
Insignia Task addition in MSBuild Task Reference

NPatch:
Added Insignia Task documentation file having used Candle as reference.
Used some descriptions for the Insignia specific properties from the insignia.cs command line help strings.
Also added a link to the insignia.html in the MSBuild Task Reference.

There might be some more space for improvement when it comes to clarifying which Insignia specific property sets are required for which mode,mode being one of four:InscribeMSI,DetachEngine,ReattachEngine,SignBundle or Engine. Also some properties might be missing. For one, RunAsSeparateProcess was omitted because I didn't understand how exactly insignia.exe could be invoked without it being a separate process.


NPatch:
Perhaps I could try to add the appropriate Insignia Tasks in ~/overview/insignia.html as well.
The guide there was very appropriate but it took a bit to figure out which properties to use for which differrent mode.

barnson:
None of the other "Tools and Concepts" pages do, so I think this is consistent. It might make sense for each of them to point to the MSBuild pages or to combine them (maybe one table with switches and task parameters?).

PR #392, opened 2017-01-24
Fix votive bug https://github.com/wixtoolset/issues/issues/5444

ddv239:
The reality is that the configName argument of the IVsBuildPropertyStorage.GetPropertyValue method is the canonical name of the configuration but the code in the Votive's implementation is using it as the config part and leaves the platform part empty. This in particular causes the invalid path exception and keeps the project file dirty in the VS IDE as result.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 [Pp]ackages/
 ipch/
 .vs/
+obj/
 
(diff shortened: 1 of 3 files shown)


barnson:
Thanks!

PR #395, opened 2017-02-06
Add support for Visual Studio 2017 (current as of 15.0.26127.3).

barnson:
Adds C++ CA template but doesn't install it because templates are per-instance.
Implements wixtoolset/issues/issues/5484.
Note that this is just build-time support, not Votive.


tools/WixBuild.props
```diff
@@ -184,7 +192,22 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <VCPath>$([MSBuild]::GetRegistryValueFromView('HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\VisualStudio\14.0\Setup\VC', 'ProductDir', null, RegistryView.Registry64, RegistryView.Registry32))</VCPath>
+    <VS2017Path Condition=" '$(VS2017Path)'=='' and '$(VsInstallRoot)'!=''">$(VsInstallRoot)\</VS2017Path>
```
heaths: Maybe beyond the scope right now, but a task can be added to find the latest instance with the `Microsoft.VisualStudio.Workload.NativeDesktop`. I hope to get around to one eventually (other projects in the works working up to it), but - like with https://github.com/Microsoft/vssetup.powersell - would effectively look like:

```powershell
$path = get-vssetupinstance | select-vssetupinstance -version '[15.0,16.0)' -require Microsoft.VisualStudio.Workload.NativeDesktop -latest | select-object -expand InstallationPath
```

tools/WixBuild.props
```diff
@@ -71,10 +71,10 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <PlatformSdkInstallPath>$([MSBuild]::GetRegistryValueFromView('HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows Kits\Installed Roots', 'KitsRoot10', null, RegistryView.Registry64, RegistryView.Registry32))</PlatformSdkInstallPath>
+    <PotentialPlatformSdkInstallPath>$([MSBuild]::GetRegistryValueFromView('HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows Kits\Installed Roots', 'KitsRoot10', null, RegistryView.Registry64, RegistryView.Registry32))</PotentialPlatformSdkInstallPath>
+    <PlatformSdkInstallPath Condition=" '$(PotentialPlatformSdkInstallPath)'!='' and Exists('$(PotentialPlatformSdkInstallPath)\include\10.0.14393.0\um') ">$(PotentialPlatformSdkInstallPath)</PlatformSdkInstallPath>
     <PlatformSdkRegistryVersion Condition=" '$(PlatformSdkInstallPath)'!='' ">v10</PlatformSdkRegistryVersion>
-    <PlatformSdkIncludeRoot Condition=" '$(PlatformSdkIncludeRoot)'=='' and Exists('$(PlatformSdkInstallPath)\include\10.0.14393.0') ">$(PlatformSdkInstallPath)\include\10.0.14393.0</PlatformSdkIncludeRoot>
```
barnson: The path hasn't been updated since Anniversary so I don't want to waste the effort to write a task. I reserve the right to change my mind if the frequency changes.

tools/WixBuild.props
barnson: I'm relying on the user to run MSBuild from whichever instance they want.

rseanhall: We can only rely on this until the next version of VS comes out.  Because then we'll only be able to detect either VS2017 or VSvNext.

barnson: True. (Well, presumably; I mean, Windows 10 is the "last" Windows, right? 😈 ) There are MSBuild properties we can rely on, assuming MSBuild versions as you'd expect.

PR #396, opened 2017-02-10
Fix #4094 - Check if current package shows internal UI before show MSI message.

aleksandr-grechko:
Fix [wixtoolset/issues#4094](https://github.com/wixtoolset/issues/issues/4094).
Adds condition to check current package shows internal UI before show MSI message like it is done in OnError handler in WixStandardBootstrapperApplication.


aleksandr-grechko:
Added changes for uninstall or repair actions because burn engine doesn't show internal UI for any value of DisplayInternalUI attribute.

barnson:
Thanks!

PR #398, opened 2017-02-10
Create history enty for Fix #4932

MarcSchaetz:


robmen:
Please add this commit to the branch with the PR for the actual fix. We don't want to separate the History entry and the fix.

PR #402, opened 2017-02-12
Correct `RemoveFolder` entry in shortcut how-to.

barnson:
Fixes wixtoolset/issues/issues#4869.


robmen:
Fixes wixtoolset/issues#4869

PR #403, opened 2017-02-12
Prevent use of SSE for old-old CPUs.

barnson:
Fixes wixtoolset/issues/issues#4876.


robmen:
Fixes wixtoolset/issues#4876

PR #404, opened 2017-02-12
Fix null-check to check the right null.

barnson:
Fixes wixtoolset/issues/issues#5318.


robmen:
Fixes wixtoolset/issues/issues#5318

PR #406, opened 2017-02-12
Update bundle splash screen with new logo.

barnson:


robmen:
Fixes wixtoolset/issues#5475

PR #407, opened 2017-02-13
Correct condition to detect missing .NET 2.0/3.5.1.

barnson:


barnson:
https://github.com/wixtoolset/issues/issues/3872

PR #409, opened 2017-02-14
Fix#3813: Wrong binding in IIS

fyodorkor:


barnson:
Too big for v3. Will take v4 PR.

PR #410, opened 2017-02-17
Fix ARM build, some paths were moved

woachk:
Fix wixarm.proj. It was unbuildable in its current form.

### DIFF
--- a/src/wixarm.proj
+++ b/src/wixarm.proj
@@ -14,19 +14,19 @@
   </ItemDefinitionGroup>
 
   <ItemGroup Condition=" $(BuildARM) ">
-    <ProjectReference Include="dutil\dutil.vcxproj" />
-    <ProjectReference Include="ca\wcautil\wcautil.vcxproj" />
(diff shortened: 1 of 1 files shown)


rseanhall:
FYI, we don't currently build ARM because Microsoft hasn't released the required tools. When I try to compile ARM today, I get `error MSB8022: Compiling Desktop applications for the ARM platform is not supported.`  If you are able to build WiX with publicly released tools, we will be happy to start building and deploying it with your help.

woachk:
@rseanhall it's an artificial limitation, See http://stackoverflow.com/questions/11151474/can-arm-desktop-programs-be-built-using-visual-studio-2012 . The ARM SDK in Win10 is complete.

woachk:
@rseanhall you can also add an environment variable ;) Since Windows 10, the ARM SDK has all the needed .libs and is safe to use. (Windows 8.1 is to a lesser extent, but enough for WiX, and avoid 8.0)

PR #411, opened 2017-02-18
Fix 5265 and 5307

robmen:
Prevent bad use of ".." in payload names in the Compiler and Binder.
Fixes wixtoolset/issues#5265

Correctly set and document SystemFolder and System64Folder in Burn
Fixes wixtoolset/issues#5307

### DIFF
--- a/src/burn/engine/variable.cpp
+++ b/src/burn/engine/variable.cpp
@@ -829,7 +829,7 @@ extern "C" HRESULT VariableSerialize(
 
         // If we aren't persisting, include only variables that aren't rejected by the elevated process.
         // If we are persisting, include only variables that should be persisted.
-        fIncluded = (!fPersisting && BURN_VARIABLE_INTERNAL_TYPE_BUILTIN != pVariable->internalType) || 
+        fIncluded = (!fPersisting && BURN_VARIABLE_INTERNAL_TYPE_BUILTIN != pVariable->internalType) ||
                     (fPersisting && pVariable->fPersisted);
 
         // Write included flag.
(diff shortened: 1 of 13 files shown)


src/burn/engine/variable.cpp
```diff
@@ -1897,26 +1897,35 @@ static HRESULT InitializeVariableSystemFolder(
     BOOL f64 = (BOOL)dwpData;
     WCHAR wzSystemFolder[MAX_PATH] = { };
 
-#ifndef _WIN64
```
rseanhall: This code used to return the right answer when running on a 32-bit OS, and swap SystemFolder with System64Folder on a 64-bit OS. Now it returns the right answer on 32-bit and 64-bit. Theoretically, someone could have been assuming that they're always deploying to a 64-bit OS (because their org's IT takes care of that) and then took a dependency on them being swapped. Are we justifying this breaking change by preferring the ability to support both 32-bit and 64-bit OS's? Should we have a warning somewhere?

robmen: I'm fine if we want to say this needs to move to WiX v4.0 because it is "a fix breaking today's behavior". Adding warning doesn't make sense to me.

rseanhall: I'm leaning towards taking it anyway.

robmen: Sounds like @rseanhall, you've come to the same conclusion I did in [my blog post](http://robmensching.com/blog/posts/2017/2/17/lets-code-wix-rooting-payloads-and-fixing-the-system-in-burn/):

> What if this fix breaks backwards compatbility. Then I remind myself this code never returns the right answer. Raymond Chen just wrote about a similar problem. Although our issue isn't a bluescreen or a crash. Hmmm... Nope, I'm still fixing it tonight.

😃

barnson: It's not likely to break typical uses and `System64Folder` wasn't documented, so I'm OK with the change but it should probably get highlighted in release notes.

rseanhall:
Please make a pull request for wix4

robmen:
Port done.

PR #412, opened 2017-02-18
Fix bug in WixBA where the buttons weren't enabled

rseanhall:
until clicking somewhere.


barnson:
So glad it wasn't (just) my imagination...

PR #414, opened 2017-02-20
Get ARM building with VS2017 and Win10 SDK

rseanhall:
I had to go to the Individual Components in the VS2017 installer and select "Visual C++ compilers and libraries for ARM".

I haven't tested it, but based on https://github.com/wixtoolset/wix3/pull/410 and https://github.com/dokan-dev/dokany/issues/431 it sounds like it should work.

### DIFF
--- a/tools/WixBuild.vcxproj.props
+++ b/tools/WixBuild.vcxproj.props
@@ -25,6 +25,7 @@
     <!-- VS2015: turn off 'typedef ignored when no variable declared' -->
     <!-- VS2015: turn off 'declaration hides class member' -->
     <DisableSpecificCompilerWarnings Condition=" $(PlatformToolset.StartsWith('v14')) ">$(DisableSpecificCompilerWarnings);4091;4458</DisableSpecificCompilerWarnings>
+    <DisableSpecificCompilerWarnings Condition=" '$(Platform)'=='arm' ">$(DisableSpecificCompilerWarnings);4996</DisableSpecificCompilerWarnings>
   </PropertyGroup>
 
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.Default.props" />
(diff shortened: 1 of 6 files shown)


tools/WixBuild.vcxproj.props
```diff
@@ -52,6 +53,8 @@
   <PropertyGroup>
     <ResourceOnlyDll Condition=" '$(ResourceOnlyDll)'=='' ">false</ResourceOnlyDll>
     <GenerateManifest Condition="$(ResourceOnlyDll)">false</GenerateManifest>
+    <WindowsSDKDesktopARMSupport Condition=" '$(Platform)'=='arm' and $(VS2017Available) ">true</WindowsSDKDesktopARMSupport>
```
barnson: Is there an easy way to check whether ARM support was installed? Better to skip than fail the build, imo.

rseanhall: I didn't auto enable building ARM, so you still have to add /p:BuildARM=true to the build command line. For this property, I should have replaced `VS2017Available` with `Exists(PlatformSdkArmLibraryRoot)` because it's about whether the WinSDK has ARM and not the build tools. I think we have to assume that the person installs the ARM support in VS for now, just like we do for XP support.

woachk:
@barnson The ARM SDK is complete on all VS2015 installs with UWP tools. There is no need to require VS2017. VS2013 and older should be prevented to work though.

tools/WixBuild.vcxproj.props
woachk: ARM .libs are still included in the Windows 10 SDK if you don't select ARM build tools

rseanhall:
I pushed support for building ARM with VS2015.

woachk:
@rseanhall thanks ;)
did you test it on your own machine or not yet?

rseanhall:
I have built ARM on my machine with VS2015 and VS2017. I don't have any ARM environments to test the results.

woachk:
@rseanhall so you can merge then :)

PR #419, opened 2017-03-06
WIXFEAT:5435 - Include the BA's directory in the DLL search path

rseanhall:
[`LoadLibraryEx`](https://msdn.microsoft.com/en-us/library/windows/desktop/ms684179(v=vs.85).aspx) with `LOAD_WITH_ALTERED_SEARCH_PATH` uses the target DLLs directory while loading it instead of the running application's directory, and is available on all platforms that we support.

Fixes wixtoolset/issues#5435.

### DIFF
--- a/history/5435.md
+++ b/history/5435.md
@@ -1,0 +1,1 @@
+* SeanHall: WIXFEAT:5435 - When loading the BA, include the BA's directory in the DLL search path.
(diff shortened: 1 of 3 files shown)


robmen:
Fantastic! Beyond fantastic. I had to go read altered search paths for the last 10 minutes to make sure this worked out. Perfect.

rseanhall:
Yeah, I also spent a fair amount of time reading up on all the different search methods they have available. When I saw the details for `LOAD_WITH_ALTERED_SEARCH_PATH`, I really hoped that it would work out. And then spent more time than I wanted working with Qt to verify that it worked.

PR #421, opened 2017-03-07
Add detection property support for VS2017

heaths:
Relates to issue wixtoolset/issues#5433 - though not a complete fix until changes for _VSIXInstaller.exe_ are implements. Tested with simple MSI that referenced the various properties supported by the `VSFindInstances` custom action.


heaths:
Adding @et13 to the review. I also realized I need to update the _chm_ and will push to this PR when complete.

src/ext/VSExtension/ca/vsca.cpp
```diff
@@ -0,0 +1,360 @@
...
+    ISetupConfiguration* pConfiguration = NULL;
+    ISetupHelper* pHelper = NULL;
+    IEnumSetupInstances* pEnumInstances = NULL;
+    ISetupInstance* rgpInstances[1] = {};
+    ISetupInstance* pInstance = NULL;
+    ULONG cInstancesFetched = 0;
+    BSTR bstrVersion = NULL;
+    DWORD64 qwVersion = 0;
+
+    hr = WcaInitialize(hInstall, "FindInstances");
```
barnson: `VSFindInstances` (CA id)

src/ext/VSExtension/ca/vsca.cpp
```diff
@@ -0,0 +1,360 @@
...
+    {
+        er = ERROR_INSTALL_FAILURE;
+    }
+
+    return WcaFinalize(er);
+}
+
+static HRESULT InstanceIsGreater(
+    __in_opt ISetupInstance* pPreviousInstance,
+    __in DWORD64 qwPreviousInstance,
```
barnson: qwVersion...?

barnson: Just found myself confused.

src/ext/VSExtension/ca/vsca.cpp
```diff
@@ -0,0 +1,360 @@
...
+    {
+        if (pLatest)
+        {
+            hr = ProcessInstance(pLatest, L"VS2017_ROOT_FOLDER", countof(rgComponents), rgComponents);
+            ExitOnFailure(hr, "Failed to process VS2017 instance.");
+        }
+    }
+    else if (pInstance)
+    {
+        hr =  InstanceIsGreater(pLatest, qwLatest, pInstance, qwVersion);
```
barnson: whitespace

barnson:
LGTM.

src/ext/VSExtension/ca/vsca.cpp
```diff
@@ -0,0 +1,360 @@
...
+    {
+        hr = pEnumInstances->Next(1, rgpInstances, &cInstancesFetched);
+        if (SUCCEEDED(hr) && cInstancesFetched)
+        {
+            pInstance = rgpInstances[0];
+            if (pInstance)
+            {
+                if (pHelper)
+                {
+                    hr = pInstance->GetInstallationVersion(&bstrVersion);
```
et13: Looks like this could overwrite bstrVersion from a previous iteration of the loop, leaking the prior string.

src/ext/VSExtension/ca/vsca.cpp
```diff
@@ -0,0 +1,360 @@
...
+                    ExitOnFailure(hr, "Failed to get installation version.");
+
+                    hr = pHelper->ParseVersion(bstrVersion, &qwVersion);
+                    ExitOnFailure(hr, "Failed to parse installation version.");
+                }
+
+                for (DWORD i = 0; i < countof(vrgInstances); ++i)
+                {
+                    const VS_INSTANCE* pElem = &vrgInstances[0];
```
et13: vrgInstances[i]

src/ext/VSExtension/ca/vsca.cpp
```diff
@@ -0,0 +1,360 @@
...
+            }
+
+            ReleaseNullObject(pInstance);
+        }
+    } while (SUCCEEDED(hr) && cInstancesFetched);
+
+    // Complete all registered processing functions.
+    for (DWORD i = 0; i < countof(vrgInstances); ++i)
+    {
+        const VS_INSTANCE* pElem = &vrgInstances[0];
```
et13: vrgInstances[i]

src/ext/VSExtension/ca/vsca.cpp
```diff
@@ -0,0 +1,360 @@
...
+        return qwPreviousVersion < qwCurrentVersion ? S_OK : S_FALSE;
+    }
+
+    hr = pPreviousInstance->GetInstallDate(&ftPreviousInstance);
+    ExitOnFailure(hr, "Failed to get previous install date.");
+
+    hr = pCurrentInstance->GetInstallDate(&ftCurrentInstance);
+    ExitOnFailure(hr, "Failed to get current install date.");
+
+    return 0 <= ::CompareFileTime(&ftPreviousInstance, &ftCurrentInstance) ? S_OK : S_FALSE;
```
et13: Should this really be '<' instead of '<='?  Seems inconsistent with the version check on line 186 which uses '<'.

et13: From the documentation for CompareFileTime, it looks like the argument order is wrong.  CompareFileTime returns 1 if the first file time (ftPreviousinstance) is later than the second time file time (ftCurrentInstance).

barnson:
Thanks!

src/ext/VSExtension/ca/vsca.cpp
heaths: Technically, `=` should never happen with install dates since we block concurrent installs but I didn't want to specifically exclude the possibility, so figured if they are equal use the "current" instance.

src/ext/VSExtension/ca/vsca.cpp
heaths: You're right. I get so used to using classes like `bstr_t` in other codebases I often forget.

PR #426, opened 2017-03-08
Adding attribute LoggingBaseFolder to Bundle's Log element

frankchdez:
This will allow placing the logs created by the burn bundle in a different folder other than %TEMP%.
If omitted it's backward compatible with existing functionality.

#### Sample xml

```
<Wix ...
    <Bundle ....
        <Log LoggingBaseFolder="[CommonAppDataFolder]\Acme" Prefix="InstallLogs\AwesomeInstaller" />
        <!-- ... -->
```

### DIFF
--- a/src/burn/engine/engine.cpp
+++ b/src/burn/engine/engine.cpp
@@ -369,6 +369,7 @@ static void UninitializeEngineState(
 
     ReleaseStr(pEngineState->log.sczExtension);
     ReleaseStr(pEngineState->log.sczPrefix);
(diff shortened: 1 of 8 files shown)


barnson:
Thanks. This won't make WiX v3.11 since we're already in RC but we'll move it to WiX v4.

gmatzen:
@frankchdez,
If the base log location cannot be opened, will the %temp% location be used?  I can foresee permission problems when changing the base log location.  The user's local %temp% should always work.

PR #428, opened 2017-03-15
Add support for nl-nl in utilextensions

hbulens:


robmen:
The .wxl won't be included in the extension if it isn't included in the libraries .wixproj. So, you will need to add a line to the project file.

PR #429, opened 2017-03-18
Prevent TargetPath from getting multiple identical items.

barnson:
wixtoolset/issues#5535


barnson:
#triage

src/tools/WixTasks/CreateProjectReferenceDefineConstants.cs
```diff
@@ -112,9 +112,12 @@ public override bool Execute()
                 if (defineConstants.ContainsKey(targetPathDefine))
                 {
                     string oldTargetPath = defineConstants[targetPathDefine];
-                    defineConstants[targetPathDefine] += ";" + targetPath;
+                    if (!targetPath.Equals(oldTargetPath, StringComparison.OrdinalIgnoreCase))
+                    {
+                        defineConstants[targetPathDefine] += ";" + targetPath;
```
rseanhall: Based on the error message, it sounds like light is never going to be able to handle multiple items separated by a semicolon. Not sure why we should ever do this, but I guess this would be the safest thing for v3.11.

PR #432, opened 2017-03-18
Fix util:InternetShortcut.

barnson:
wixtoolset/issues#5522


src/Setup/CoreMsi/CoreMsi.wxs
```diff
@@ -41,6 +41,11 @@
                 <RemoveFolder Id="CleanupShortcutFolder" Directory="ShortcutFolder" On="uninstall" />
             </Component>
 
+            <Component Directory="BinFolder">
+                <File Source="common\wixtoolset.org.ico" />
+                <util:InternetShortcut Id="wixtoolset.org" Directory="ShortcutFolder" Name="WiX Home Page" Target="http://wixtoolset.org/" IconFile="file://[#wixtoolset.org.ico]" />
```
rseanhall: That's one way of writing a regression test...

barnson: Exactly. It's why I wish we used WixStdBA but then we'd lose MBA "coverage."

PR #441, opened 2017-04-14
Prevent repairing of .NET prerequisite package

dvorn:
Fixes #5561 https://github.com/wixtoolset/issues/issues/5561


robmen:
Fixed in v4

PR #446, opened 2017-06-01
Issue #5601: Support RemotePayload for and Payload (not just ExePackage)

nirbar:
Done to reduce bind time of multitude unchanging payloads (e.g. SQL server). 
Without this, each build computes hashes for them all which consumes a lot of time.
Runtime behavior is unchanged.


barnson:
Issue was v4.x.

PR #447, opened 2017-06-02
Some LEADTools dlls create bad default values for some keys that incl…

cessna048:
…ude null characters and other garbage.  Truncate REG_SZ values at the first null character.


cessna048:
I am the originator of bug 3096 (formerly bug 2496 on SourceForge) I finally got around to debugging it and determined that the LEADTools filters are creating bad registry entries that have null characters and other garbage.   The .NET Framework RegistryKey API is happily reading in the entire buffer but the serializer is later choking on the nulls.   This code truncates the string at the first null character.

barnson:
Truncation is bad (data loss) but maybe in v4 we can use `[~]`?

rseanhall:
https://github.com/wixtoolset/issues/issues/3096

PR #450, opened 2017-08-23
Support for the VS 2017 VSIX installer.

dosvidos:
With this patch, one can use [VsixPackage ](http://wixtoolset.org/documentation/manual/v3/xsd/vs/vsixpackage.html) to install VSIX v3 package into all or SKU-specific VS 2017 instances.

### DIFF
--- a/src/ext/VSExtension/wixlib/VsixPackage.wxs
+++ b/src/ext/VSExtension/wixlib/VsixPackage.wxs
@@ -22,11 +22,17 @@
             </RegistrySearch>
         </Property>
 
-        <!-- VS2015 is the latest VSIX at this point in time, so search for that by default. -->
         <Property Id="VS2015_VSIX_INSTALLER_PATH" Secure="yes">
             <RegistrySearch Id="WixVS2015EnvDir" Root="HKLM" Key="SOFTWARE\Microsoft\VisualStudio\14.0\Setup\VS" Name="EnvironmentDirectory" Type="directory">
                 <FileSearch Id="WixVS2015VsixInstaller" Name="VSIXInstaller.exe" />
             </RegistrySearch>
+        </Property>
+
+        <!-- Currently, VS2017 contains the latest VSIX installer, so search it by default. -->
+        <Property Id="VS2017_VSIX_INSTALLER_PATH" Secure="yes">
+            <DirectorySearchRef Id="VS2017EnvironmentDirectorySearch" Parent="VS2017DirectorySearch" Path="Common7\IDE">
+                <FileSearch Id="WixVS2017VsixInstallerPath" Name="VSIXInstaller.exe" />
+            </DirectorySearchRef>
         </Property>
 
         <Property Id="VS_VSIX_INSTALLER_PATH" Secure="yes" />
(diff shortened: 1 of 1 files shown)


src/ext/VSExtension/wixlib/VsixPackage.wxs
```diff
@@ -22,13 +22,19 @@
...
         <Property Id="VS2015_VSIX_INSTALLER_PATH" Secure="yes">
             <RegistrySearch Id="WixVS2015EnvDir" Root="HKLM" Key="SOFTWARE\Microsoft\VisualStudio\14.0\Setup\VS" Name="EnvironmentDirectory" Type="directory">
                 <FileSearch Id="WixVS2015VsixInstaller" Name="VSIXInstaller.exe" />
             </RegistrySearch>
         </Property>
 
+        <!-- Currently, VS2017 contains the latest VSIX installer, so search it by default. -->
+        <Property Id="VS2017_VSIX_INSTALLER_PATH" Secure="yes">
+            <DirectorySearchRef Id="VS2017EnvironmentDirectorySearch" Parent="VS2017DirectorySearch" Path="Common7\IDE">
+                <FileSearch Id="WixVS2017VsixInstallerPath" Name="VSIXInstaller.exe" />
```
heaths: We recommend using the VSIXInstaller in the search engine root. We update that one with the installer so it's always newer and may eventually remove the one installed with VS.

Since I set the VS2017 property to the latest version, this would be a fine backup, but the installer party is already set by the custom actions.

The real problem is that if a VSIX triggers an MSI install, the install will fail. There's a WIP on good you should read if you want to improve this feature.

src/ext/VSExtension/wixlib/VsixPackage.wxs
```diff
@@ -47,7 +53,9 @@
         </Property>
 
         <!-- Use the latest VS- or VWD-installed VSIXInstaller.exe. -->
-        <SetProperty Action="SetVS2015Vsix" Id="VS_VSIX_INSTALLER_PATH" Value="[VS2015_VSIX_INSTALLER_PATH]" Sequence="both" After="AppSearch">NOT VS_VSIX_INSTALLER_PATH AND VS2015_VSIX_INSTALLER_PATH</SetProperty>
+        <SetProperty Action="SetVS2017Vsix" Id="VS_VSIX_INSTALLER_PATH" Value="[VS2017_VSIX_INSTALLER_PATH]" Sequence="both" After="AppSearch">NOT VS_VSIX_INSTALLER_PATH AND VS2017_VSIX_INSTALLER_PATH</SetProperty>
```
heaths: There is a custom action that already does this correctly.

src/ext/VSExtension/wixlib/VsixPackage.wxs
dosvidos: Would you kindly elaborate what is that search engine root and there it is defined?
I suppose it's %ProgramFiles(x86)%\Microsoft Visual Studio\Installer but I haven't seen this path being used in wix3 sources.

barnson: I assume you meant "setup engine," unless the Bing team is now doing VS setup...

Interestingly, with my install of 15.3.1, I have only these VSIXInstaller.exes:

* C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\Common7\IDE\VSIXInstaller.exe
* C:\Program Files (x86)\Microsoft Visual Studio\Installer\resources\app\ServiceHub\Services\Microsoft.VisualStudio.Setup.Service\VSIXInstaller.exe

barnson:
We intentionally didn't do this in WiX v3.11 because of Heath's last bullet: It's not reliable to use VSIXInstaller.exe in a custom action.

src/ext/VSExtension/wixlib/VsixPackage.wxs
dosvidos: In the latest update of VS 2017 (now I have 15.3.2) the VSIX installer is still sitting in Common7\IDE

dosvidos:
Well, I've read about the nested MSI installs as described in http://wixtoolset.org/development/wips/5433-add-support-to-detect-and-install-vsix-packages-into-vs15/
So it means that now there's literally no reliable way to install VSIX v3 from MSI.
That's just awfully inconvient.

barnson:
Agreed.

heaths:
Since VSIX v3 can do more than v2s, is there something you still need MSIs for? Also, you might consider using https://github.com/Microsoft/vsixbootstrapper so you can chain your VSIX into Burn directly. Unfortunately, there's currently no detection for VSIX v3s so anytime your bundle runs the VSIX will be reinstalled but would effectively be a no-op for any instances where it is installed.

dosvidos:
@heaths
>Since VSIX v3 can do more than v2s, is there something you still need MSIs for?
Sure - for example, MSI can be used to argument legacy MSBuild installations with a new platforms and build scripts.
In general, even with VSIX v3 the list of writable locations is limited to a few VS subdirectories.

heaths:
@dosvidos so for legacy support. Understandable.

As to the locations you want to write under the `[InstallDir]`, could you list which ones? We're looking for feedback to consider which directories would be desired for expanded support.

barnson:
This doesn't solve the difficult and probably unsolvable problem of VSIX installation via MSI.

PR #452, opened 2017-08-30
Fix for pathLengths, Issue 5314

qlik-pwe:
This is a fix for issue #5314.  It allows users to bypass the issue when calling the utils directly via command line.  When called from wix.dll, however users will have to set  RunWixToolsOutOfProc to true in their .wixproj files, at least for now.  
error LGHT0103 : The system cannot find the file #5314


barnson:
This is insufficient to enable long paths, which would include old components like the CAB API.

PR #455, opened 2017-10-19
update localization for korean

zacronan:
Korean localization file has been enhanced.


barnson:
The mbapreq.wxl doesn't need those strings.

PR #456, opened 2017-11-15
When elevated place clean room in system Temp folder

robmen:
To prevent DLL hijacking the clean room process when launched elevated,
the system Temp folder will be used instead of the user's temp folder.
This ensures the user cannot slip malicious DLLs into the clean room.

Fixes wixtoolset/issues#5724


rseanhall:
Looks good to me. At least this one is simple to fix.

PR #457, opened 2017-11-18
Fix period conistency in dialog description strings by appending a period (issue 5317)

bergmeister:
To resolve issue [5317](https://github.com/wixtoolset/issues/issues/5317), which is about some dialog description strings not ending in a period (the majority do).
A period was appended to the following variable localisations to make them consistent (note that a some of those localisations already had a period)

- BrowseDlgDescription
- InstallScopeDlgDescription
- LicenseAgreementDlgDescription
- SetupTypeDlgDescription


barnson:
We decided it's probably better to make them consistent by removing periods. I assigned the bug to myself to investigate the SDK style guide to see if there's guidance either way.

robmen:
Consider moving to WiX v5 (when development starts).

PR #458, opened 2017-12-13
Add option to suppress hard links

NeoAdonis:
This option would force WiX toolset to copy the output instead of just creating hard links to the built components.

A use-case where this is useful is in a build system where there's a file cache in place, so the output (without the suppression, the hard link) is saved on it. This means that when a component is rebuilt, the file projected by those hard links seem to be edited as well, poisoning the cache.


barnson:
Not a candidate for WiX v3.14. It can be accomplished with an extension. We'd consider some approach to handle this in WiX v4.0.

robmen:
Fixed in v4

PR #462, opened 2018-04-03
Fix for SqlExtension Incompatible with TLS 1.2 #5543

DRISHTI271110:
# WiX v3.x pull requests

For the issue: https://github.com/wixtoolset/issues/issues/5543

SqlExtension Incompatible with TLS 1.2 #5543
Here is the overview of how I am planning to fix it.

Root Cause and Fix
SQLOLEDB does not have TLS1.2 support, so need to use SQL native Client for it.

In the SqlConnectDatabase() API in file ..\wix3\src\libs\dutil\sqlutil.cpp

CoCreateInstance is being called with CLSID_SQLOLEDB as the first argument.
                                hr = ::CoCreateInstance(CLSID_SQLOLEDB, NULL, CLSCTX_INPROC_SERVER,  IID_IDBInitialize, (LPVOID*)&pidbInitialize);

Fix would be to call it with SQLNCLI_CLSID and if that fails, then as a fallback call with CLSID_SQLOLEDB.

OLD CODE:

                //obtain access to the SQLOLEDB provider
                hr = ::CoCreateInstance(CLSID_SQLOLEDB, NULL, CLSCTX_INPROC_SERVER,
                                IID_IDBInitialize, (LPVOID*)&pidbInitialize);
                ExitOnFailure(hr, "failed to create IID_IDBInitialize object");

NEW CODE:

                //obtain access to the SQLOLEDB provider
                hr = ::CoCreateInstance(SQLNCLI_CLSID, NULL, CLSCTX_INPROC_SERVER,
                                IID_IDBInitialize, (LPVOID*)&pidbInitialize);
                if (FAILED(hr))
                {
                hr = ::CoCreateInstance(CLSID_SQLOLEDB, NULL, CLSCTX_INPROC_SERVER,
                                                IID_IDBInitialize, (LPVOID*)&pidbInitialize);
                }
                ExitOnFailure(hr, "failed to create IID_IDBInitialize object");


barnson:
This needs to be in wix4 repo. We're not taking features in wix3 any more.

DRISHTI271110:
We need this in WIX3.
Moreover this is not a feature as such, it is like a bug fix with very less minimal code changes.

qianchen0613:
@barnson could you consider add this into Wix3.11 stable version? I totally understand our future would be wix 4.0, but TLS 1.2 support issue becomes a serious showstopper for WIX 3.11 as TLS 1.0 and 1.1 is end of life. All SQL installation will fail and there are no workarounds.

PR #464, opened 2018-04-26
Fix DependencyDetectProviderKeyPackageId to return on first GetProvid…

MattHyman:
DependencyDetectProviderKeyPackageId was incorrectly refactored in a448a60ec8fe428058cc3ac01aeae1175b1d6402 to no longer return success after finding the first provider key. This results in incorrect provider behavior if there are multiple provider keys and the target one is not last.

This PR fixes the issue.


MattHyman:
I was not aware of the previous fix in wix4 when I wrote this, I am willing to switch to that solution if we feel it is the right thing to do.

MattHyman:
Switching to a cherry-pick of wixtoolset/wix4#218.

PR #469, opened 2018-08-03
Fix #5859 (Melt crashes melting MSIs with elevated privileges)

dtuchlinsky:
Fix bug where melt.exe crashes when attempting to write to a temporary file in a temporary directory that no longer exists.  TempFileCollection will create a sub-directory in %TMP% when run with elevated UAC privileges and when the instance of TempFileCollection is deleted the sub-directory is also deleted.

### DIFF
--- a/src/tools/wix/Pdb.cs
+++ b/src/tools/wix/Pdb.cs
@@ -84,6 +84,7 @@ namespace Microsoft.Tools.WindowsInstallerXml
         internal static Pdb Load(Stream stream, Uri uri, bool suppressVersionCheck, bool suppressSchema)
         {
             XmlReader reader = null;
+            TempFileCollection tempFileCollection = null;
             string cabPath = null;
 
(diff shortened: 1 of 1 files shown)


barnson:
wixtoolset/issues#5859

src/tools/wix/Pdb.cs
```diff
@@ -84,17 +84,16 @@ public static Pdb Load(string path, bool suppressVersionCheck, bool suppressSche
         internal static Pdb Load(Stream stream, Uri uri, bool suppressVersionCheck, bool suppressSchema)
         {
             XmlReader reader = null;
+            TempFileCollection tempFileCollection = null;
```
BMurri: Not critical, but I would recommend a  try/finally wherein you dispose of the TempFileCollection if it is not assigned to pdb.Output.TempFiles

chzawist:
Thanks for providing this patch 👍

PR #473, opened 2018-08-21
[wix3] Display more specific error message when UpdateFileRow cannot access a file

jozefizso:
Is this a good approach for fixing https://github.com/wixtoolset/issues/issues/5417?


barnson:
The code change is fine but the message should be in the error class, not verboses. And it could probably just reuse the `WixErrors.Win32Exception` message.

PR #474, opened 2018-08-24
Localize missing entries in Slovak translation

jozefizso:
Translated two new entries (**NET452WIN7RTMErrorMessage** and **ErrorFailNoActionReboot**) into Slovak localization file.


barnson:
Thanks!

PR #475, opened 2018-08-31
Remove incorrect Votive support

barnson:


rseanhall:
I thought the new extension did support VS2010.

barnson:
It does. But if we missed updating this page for VS2012, VS2015, and VS2017, I figure it's safer to not list them here...

PR #476, opened 2018-12-10
WIXBUG:5711 - Remove last reference to .NET 2.0 assembly

rseanhall:
Reference Microsoft.Build.Utilities.v4.0 instead of Microsoft.Build.Utilities.

Fixes https://github.com/wixtoolset/issues/issues/5711


barnson:
:shipit:

PR #477, opened 2018-12-18
WIXFEAT:5658 - Retry elevation once if we think it failed due to AV

rseanhall:
v3 version of https://github.com/wixtoolset/wix4/pull/262 in case we want it here.

### DIFF
--- a/src/burn/engine/core.cpp
+++ b/src/burn/engine/core.cpp
@@ -514,10 +514,11 @@ extern "C" HRESULT CoreElevate(
     )
 {
     HRESULT hr = S_OK;
+    DWORD cAVRetryAttempts = 0;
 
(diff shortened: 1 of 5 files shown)


src/burn/engine/core.cpp
```diff
@@ -514,17 +514,23 @@ extern "C" HRESULT CoreElevate(
     )
 {
     HRESULT hr = S_OK;
+    DWORD cAVRetryAttempts = 0;
```
BMurri: Same feedback as with wix4

DanSmith:
Works for me. 👍

rseanhall:
Superseded by https://github.com/wixtoolset/wix3/pull/504

PR #478, opened 2018-12-20
VSExtension support VS 2019 #5923

chrpai:
WiX v3 version of the PR for VS2019 support in the VSExt.   PR also submitted for v4.


chrpai:
Should I close this pull request?  It sounds like Bob doesn't want to give VS2019 support to 3.14.   I hope 4.0 is all that and a bag of chips or I'll be staying on a private build of "WiX Classic 3.12"  for a long time.

PR #481, opened 2019-03-18
Sign binaries redistrubted with customers' bundles

heaths:
Fixes wixtoolset/issues#5329. Only managed and native binaries that would ship within bundles are signed when requested.

### DIFF
--- a/src/ext/BalExtension/mba/core/core.csproj
+++ b/src/ext/BalExtension/mba/core/core.csproj
@@ -14,7 +14,7 @@
     <CreateDocumentationFile>true</CreateDocumentationFile>
     <CodeAnalysisTreatWarningsAsErrors>true</CodeAnalysisTreatWarningsAsErrors>
     <FxCopEnabled>true</FxCopEnabled>
-    <SignOutput Condition="'$(PleaseSignOutput)'!=''">true</SignOutput>
+    <ShouldSignOutput>true</ShouldSignOutput>
   </PropertyGroup>
 
   <ItemGroup>
--- a/src/ext/BalExtension/mba/core/core.csproj
+++ b/src/ext/BalExtension/mba/core/core.csproj
@@ -14,6 +14,7 @@
     <CreateDocumentationFile>true</CreateDocumentationFile>
     <CodeAnalysisTreatWarningsAsErrors>true</CodeAnalysisTreatWarningsAsErrors>
     <FxCopEnabled>true</FxCopEnabled>
+    <SignOutput Condition="'$(PleaseSignOutput)'!=''">true</SignOutput>
   </PropertyGroup>
 
   <ItemGroup>
--- a/tools/WixBuild.wixproj.targets
+++ b/tools/WixBuild.wixproj.targets
@@ -3,40 +3,7 @@
 
 
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003" ToolsVersion="4.0">
-  <PropertyGroup>
-    <SignToolExe>&quot;$(PlatformSdkBinPath)signtool.exe&quot;</SignToolExe>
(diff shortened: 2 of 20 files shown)


src/ext/BalExtension/mba/core/core.csproj
```diff
@@ -14,6 +14,7 @@
     <CreateDocumentationFile>true</CreateDocumentationFile>
     <CodeAnalysisTreatWarningsAsErrors>true</CodeAnalysisTreatWarningsAsErrors>
     <FxCopEnabled>true</FxCopEnabled>
+    <SignOutput Condition="'$(PleaseSignOutput)'!=''">true</SignOutput>
```
robmen: Can this be centralized to the Tools\WixBuild.csproj.props/targets?

heaths: This is the only managed DLL that may be redistributed. Given the perf hit of Authenticode signing managed assemblies, do you want them all to be signed? In that case, should be we signing all EXEs and DLLs under \src as well as all MSIs and not just the bundle? Per the PR description, I was intending to keep this change to just redistributed binaries. I imagine not too many build machines could successfully use Device Guard given the very heterogenous build systems in play these days.

tools/WixBuild.wixproj.targets
```diff
@@ -3,41 +3,8 @@
 
 
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003" ToolsVersion="4.0">
-  <PropertyGroup>
```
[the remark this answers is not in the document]
heaths: No. WixBuild.wixproj.props already has:

```xml
    <SignOutput Condition=" '$(OutputType)'=='Bundle' and '$(PleaseSignOutput)'!='' ">true</SignOutput>
```
Comments in the old WixBuild.wixproj.props indicated that MSIs and CABs wouldn't be signed since the bundle was, so that is likely why. Now, if we want to sign more things (perhaps all the things under /src except, perhaps, DTF), I could centralize it accordingly.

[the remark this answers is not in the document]
barnson: :+1:

src/ext/BalExtension/mba/core/core.csproj
robmen: Ahh, that makes sense. The other managed DLLs that get redistributed are DTF, in a managed custom action. We probably need to sign those as well then.

It is weird to spread `PleaseSignOutput` around the code (it was meant to be hidden away in the infrastructure) so maybe for managed code projects we want signed, we should add a property like `Signable` or `ShouldSign` (or some better word than I'm coming up with right now) and move setting the `SignOutput` property to the infrastructure. That would allow `PleaseSignOutput` to be hidden again...

BMurri:
Hey! I don't mean to stop the presses, but I don't see src/ext/ca/ in the list. Those absolutely end up in customer bundles (via the MSIs, to be sure, but...)

heaths:
@BMurri, see https://github.com/wixtoolset/wix3/pull/481/files#diff-6ec57e3cf1fa4f7a7ba0393b4ec42679R51. All native projects under src/ext get signed.

BMurri:
@heaths /tools/appsettings.json affects /src/ext? The connection isn't obvious.

heaths:
What? No, the WixBuild.vcxproj.props does, which is consistent with other WixBuild.*.props/targets and global per-project type settings.

heaths:
Have you had a chance to try this with the new values you were given for real-signing? You can set them as environment variables (or command line properties) using the `SignClientUser` and `SignClientSecret` variables.

robmen:
Completing this PR is the second thing on my list to do today. So, if it doesn't happen today, it'll be next.

PR #482, opened 2019-04-07
Make sure Light task's output matches OutputFile

pascalpfeil:
I have been playing with changing light's `OutputFile` and the console output did not quite match. I created a custom Property similar to `PdbOutputFile` so that console output always matches the `OutputFile`.


robmen:
Consider reimplementing in WiX v5 if WiX v4 not sufficient.

PR #483, opened 2019-04-09
Release taskbar before COM uninitialized to prevent crash

robmen:
Expected fix for wixtoolset/issues#5966


rseanhall:
This looks good to me. Are we guessing that it fixes wixtoolset/issues#5966 or were you guys able to verify this fixes it?

barnson:
I was able to verify with Nikola that the fix fixes the problem in a test bundle that uses `RebootPending`.

PR #486, opened 2019-05-03
Add support for Visual Studio 2019

heaths:
Fixes wixtoolset/issues#5979. Ran end-to-end tests on machines with both VS2017 and VS2019, and only VS2019 with test MSI before and after changes. Properties for all versions on both machines were defined correctly after changes.


barnson:
:100:

PR #487, opened 2019-05-09
wixstdba: Add auto-launch option via 'AutoLaunch' variable.

nickdademo:
Any feedback welcome on this PR.

A feature often requested that could be quite useful, e.g:
 * https://stackoverflow.com/questions/39745742/how-to-launch-the-application-after-a-quiet-call-to-wix-burn-bootstrap
* http://windows-installer-xml-wix-toolset.687559.n2.nabble.com/How-to-force-to-launch-application-in-silent-install-mode-td4021727.html
* https://stackoverflow.com/questions/12823722/start-application-after-installation-using-wix-burn
 * https://github.com/vslavik/winsparkle/issues/132

If accepted, perhaps WiX4 would be a better place for it?

### Within bootstrapper .wxs Bundle:
```
<Variable Name="AutoLaunch" bal:Overridable="yes" Type="numeric" Value="0" />
```
### When running bootstrapper:
```
BootstrapperInstaller.exe AutoLaunch=1
```

Auto-launching will **not** be performed if a restart is to be performed.

### DIFF
--- a/src/ext/BalExtension/wixstdba/WixStandardBootstrapperApplication.cpp
+++ b/src/ext/BalExtension/wixstdba/WixStandardBootstrapperApplication.cpp
@@ -965,19 +965,24 @@ public: // IBootstrapperApplication
         SetState(WIXSTDBA_STATE_APPLIED, hrStatus);
         SetTaskbarButtonProgress(100); // show full progress bar, green, yellow, or red
 
-        // Auto-launch if configured to do so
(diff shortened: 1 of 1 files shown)


barnson:
This should go into WiX v4 first but should start with a thread on wix-devs: Triage was concerned about the name `AutoLaunch` as not being clearly to control this feature.

nickdademo:
@barnson Do I need to create this thread on wix-devs? Or is it something the WiX team will do?

robmen:
Consider moving to WiX v5 (when development starts).

PR #492, opened 2019-07-24
Add support for FIPS compliance to Heat.exe and GenerateCompileWithObjectPath Task

Herohtar:
# Fixes https://github.com/wixtoolset/issues/issues/4368
- `heat.exe` is now supports the `-fips` command line switch and will use SHA1 for ID generation when present, in the same manner as `candle.exe`
- Heat build tasks now have a `FipsCompliant` property
- Heat targets have been updated to pass the value of the `<FipsCompliant>` tag to the build tasks
- Change is backwards compatible by defaulting to `false` when nothing is provided (opt-in)

# Fixes https://github.com/wixtoolset/issues/issues/5672
- GenerateCompileWithObjectPath uses SHA1 instead of MD5 for ID generation


Herohtar:
Per comments in WiX Online Meeting <span>#</span>174, there is no need for backward compatibility in `GenerateCompileWithObjectPaths` since it is only used for intermediate files in the `obj` directory. I've removed the flags and simply replaced MD5 with SHA1.

perumalsamya:
Hi @Herohtar, I am still facing this issue.. 

Could you please mention which version of WIX contains this fix ?


Thanks

Herohtar:
@perumalsamya The WiX team has only included it in the [3.14 dev builds](https://wixtoolset.org/releases/development/).

perumalsamya:
@Herohtar Thank you.. It helps me a lot..

PR #493, opened 2019-08-12
WIXBUG:5286 - Fix the issue 'Burn: Prerequisite bootstrapper applicat…

DemiDOS:
Changed determining when restart is needed and when it is allowed.

### DIFF
--- a/src/ext/BalExtension/wixstdba/WixStandardBootstrapperApplication.cpp
+++ b/src/ext/BalExtension/wixstdba/WixStandardBootstrapperApplication.cpp
@@ -927,9 +927,9 @@ public: // IBootstrapperApplication
         m_restartResult = restart; // remember the restart result so we return the correct error code no matter what the user chooses to do in the UI.
 
         // If a restart was encountered and we are not suppressing restarts, then restart is required.
-        m_fRestartRequired = (BOOTSTRAPPER_APPLY_RESTART_NONE != restart && BOOTSTRAPPER_RESTART_NEVER < m_command.restart);
+        m_fRestartRequired = BOOTSTRAPPER_APPLY_RESTART_NONE != restart || (!m_fPrereq && BOOTSTRAPPER_RESTART_ALWAYS == m_command.restart);
         // If a restart is required and we're not displaying a UI or we are not supposed to prompt for restart then allow the restart.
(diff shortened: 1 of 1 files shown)


rseanhall:
Thanks for looking into this. Can you explain how this makes sense to the user? As far as I can tell, this just silently closes on success. So if the user walked away, then they'll come back and wonder what happened. The application that they were trying to install won't be there. They'll have to just know to restart the computer and try to install again.

DemiDOS:
Hello @rseanhall ,

this fix just prevents endless loop in case when .NET Framework is not installed. Installer was trying to install .NET Framework and after that asked the system reboot again. And installer did it (install .NET Framework and system reboot) in the endless loop. There were inaccuracies in conditions for exiting from the endless loop, which were fixed.

We use the fork of Wix Toolset in our installers, and the suggested change really fixed the original defect. We successfully issued at least one release with this change, the defect went away in UI and silent mode.

rseanhall:
@DemiDOS I never questioned whether this fixed the endless loop. I was questioning the user experience of silently closing during full UI without telling the user a restart is required.

rseanhall:
@DemiDOS I took another look at this when fixing the issue in v4 and figured out that I was not reading the code correctly. With full UI and your changes in this PR, the message box [here](https://github.com/wixtoolset/wix3/blob/38c8293f766f2708b2ffd4983abe9d930153c8a1/src/ext/BalExtension/wixstdba/WixStandardBootstrapperApplication.cpp#L2462) should still be shown. However, you changed the meaning of `m_fRestartRequired` and `m_fAllowRestart` even though the bug was just about choosing to reload the BA. That's too risky for me.

In addition, you changed the behavior when in non-prereq mode and the user specifies to always restart from the command line. In the previous code, the BA would only request restart if a restart was needed. With your changes and not in full UI, it will always request a restart even if it wasn't needed. With your changes and in full UI, it looks to me like it won't prompt the user to restart so it won't restart.

I think my small commit at https://github.com/wixtoolset/wix4/commit/6895a3e9625afe647a94f970103bdbb4f58fb0f8 is a more correct and less risky fix. If you're willing to update this PR to use that change instead and test it out in your private build then we might take it in v3.14.

robmen:
Fixed in v4

PR #502, opened 2019-12-10
Feature/display internal ui

dhoerner2412:
Fixes [#3703](https://github.com/wixtoolset/issues/issues/3703)

## WiX v3.x pull requests

Since there is no stable release version of Wix 4.x, I changed the sources in the 3.x repository, because hopefully there will be another stable version of Wix 3.x

There are only a few lines of code which needs to be changed to fix this issue / implement this feature. It is now possible to change the "DisplayInternalUi" flag of each (MSI) package of a bootstrapper burn setup during the execution-process. So the end-user can decide to show the internal UI of a package durring installation and the developer must not provide two different bootstrapper setups (one with and one without GUI).

[issue]: https://github.com/wixtoolset/issues/issues/3703

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,3 +1,9 @@
+## WixBuild: Version 3.11.2.4516
+
+* HeathS: Add support for .NET Foundation signing service
+
+* RobMen: WIXBUG:6075 - Fix "Zip Slip" vulnerability in DTF.
+
 ## WixBuild: Version 3.11.1.2318
 
 * RobMen - WIXBUG:5724 - fix DLL hijack of clean room when bundle launched elevated.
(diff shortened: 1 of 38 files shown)


rseanhall:
Adding a parameter to a method in `IBootstrapperApplication` is a breaking change, so this can't be done in v3.

barnson:
Should be part of planning, not execution.

rseanhall:
Thanks, but we can't take this breaking change. This has been implemented in v4 through https://github.com/wixtoolset/issues/issues/6164. If you are still interested in implementing this in v3 then we would need to discuss it on [wix-devs](https://wixtoolset.org/documentation/mailinglist/).

PR #503, opened 2020-02-24
ARM64 support in core toolset and extensions.

barnson:


tools/WixBuild.props
```diff
@@ -166,8 +172,10 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <BuildARM>$(WindowsSDKDesktopARMSupport)</BuildARM>
+    <BuildARM Condition=" Exists('$(PlatformSdkArmLibraryRoot)') ">true</BuildARM>
     <BuildARM Condition=" '$(BuildARM)'!='true' ">false</BuildARM>
+    <BuildARM64 Condition=" Exists('$(PlatformSdkArmLibraryRoot)') ">true</BuildARM64>
+    <BuildARM64 Condition=" '$(BuildARM64)'!='true' ">false</BuildARM64>
```
rseanhall: I didn't see anything using this as a condition to skip building, did I miss something?

tools/WixBuild.Tools.targets
```diff
@@ -133,6 +133,11 @@
       Condition=" '$(PleaseSignOutput)'!='' and ('$(SignClientUser)'=='' or '$(SignClientSecret)'=='') "
       Text="Signing is requested but one or both required properites SignClientUser and SignClientSecret were not specified on the command line or as environment variables." />
 
+    <Error
+      Code="WIXBUILD016"
+      Condition=" $(BuildARM64) and !(Exists('$(PlatformSdkArm64LibraryRoot)')) "
+      Text="No suitable Windows SDK found to build ARM64" />
+
```
rseanhall: Are we going to force `BuildARM` and/or `BuildARM64` when building in release mode?

tools/WixBuild.props
barnson: Nope. Probably should have a couple of those...

tools/WixBuild.Tools.targets
barnson: Well, it'll fail when it tries to build setup...

rseanhall: But if you put `BuildARM64` in the right places, it shouldn't fail...

barnson: Yep, working on it.

tools/WixBuild.props
barnson: Ugh. Detection sucks. We need ARM64 in both the Windows SDK (included by default) and VS2017. We'd need VSWhere to determine if the ARM64 tools and libs are installed. We could also require a property to opt-in to the build, so you're not surprised when it fails on ARM/ARM64. @robmen, @rseanhall, opinions?

rseanhall: I think requiring a property to opt-in is perfectly fine.

robmen: I'm good with opt-in property as long as the OFFICIAL_BUILD fails if the opt-in property is not set. That way we don't forget to set in in the official build (because I undoubtedly will).

rseanhall:
https://github.com/wixtoolset/issues/issues/5558

rseanhall:
https://github.com/wixtoolset/issues/issues/4826

PR #510, opened 2020-03-18
Ftr multiple attached containers

nirbar:
Added support for multiple attached containers in bootstrappers.
Tested to work with larger than 2GB payloads that otherwise would have failed build
Tested for signed and unsigned bootstrappers
See https://github.com/wixtoolset/issues/issues/6144


barnson:
We're not sure about taking this feature in v3.14. But it's big enough that it would need to go into v4 first.

robmen:
Fixed in v4

PR #511, opened 2020-03-25
Issue-4875: Explicitly request RW permissions

JasonStephenson:
Deferred custom actions need to explicitly request RW permissions for the Group Policy and Enterprise physical stores.

Fixes #4875


rseanhall:
Thanks!

PR #526, opened 2021-02-05
Add support for .NET 4.8 to OneTimeWixBuildInitialization.proj

drolevar:


robmen:
Please note that we are not taking additional changes into WiX v3 at this time. Please consider contributing to WiX v4 which is under active development.

PR #527, opened 2021-02-09
WIXBUG4931 Fix drawing of image static controls

drolevar:
Fixes issue https://github.com/wixtoolset/issues/issues/4931.

This fix has been accepted in to wix4 branch.
Since it's been an annoying issue for a lot of people it might be a good idea to accept it into wix3.


rseanhall:
Thanks! As we discussed during triage, there is no regular schedule for wix3 builds. The next one *might* be as part of releasing v4 preview.0.

PR #529, opened 2021-03-08
Port dpi scaling

drolevar:
@rseanhall I don't mean it to be merged as-is, this is more to bring your attention to two commits:
41ead98 and c591867.

As for the GDI+ conversion, the main reason was the quality of upscaling.
Here's the existing `StretchBlt` implementation:
image
Here's with SetStretchBltMode(hdc, HALFTONE):
image
And here's with GDI+ stretching:
image
It's quite obvious that GDI+ is a clear winner, and it's used anyway for reading bitmaps. So why don't use it directly? I don't see any speed drawbacks either.

As for the second commit c591867, I've noticed the original size for the image was incorrect in DrawImage when Width is specified as -1 in the theme XML file.
Again, my implementation is probably not ideal for general use, it is just a suggestion.


rseanhall:
The rule is that everything must be done in v4 first, and then if someone wants it in v3 we'll consider it. We already decided this was too big for v3 in [meeting 193](https://www.firegiant.com/blog/2020/7/22/wix-online-meeting-193-highlights/).

We're not using GDI+ in the Burn engine because that would require a security review for the DLL hijacking issue. If you want it for thmutil, then please create an issue.

If you think there's a problem with `DrawImage`, then please create an issue.

rseanhall:
@drolevar I hope I didn't discourage you about this. This just wasn't the right place to suggest improvements. We create issues first, because that helps other people see what we are working on and what was included in a release. It helps people when they can find information when they run into the same problem. And it helps make sure we understand the problem being solved before committing the code that solves it.

drolevar:
@rseanhall It's OK. It was just FYI, in case you'd think it might be interesting to port it to v4. Unfortunately, I'm too busy for this right now.

PR #531, opened 2021-03-19
Add null check so we don't crash if we don't know the package id.

barnson:
Does not apply to WiX v4.

### DIFF
--- a/src/ext/BalExtension/wixstdba/WixStandardBootstrapperApplication.cpp
+++ b/src/ext/BalExtension/wixstdba/WixStandardBootstrapperApplication.cpp
@@ -807,15 +807,15 @@ public: // IBootstrapperApplication
                 }
 
                 wz = sczFormattedString ? sczFormattedString : pPackage->sczDisplayName ? pPackage->sczDisplayName : wzPackageId;
-            }
 
-            // Needs to match MsiEngineCalculateInstallUiLevel in msiengine.cpp in Burn.
-            BOOTSTRAPPER_ACTION_STATE packageAction = fExecute ? pPackage->executeAction : pPackage->rollbackAction;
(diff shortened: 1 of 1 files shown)


src/ext/BalExtension/wixstdba/WixStandardBootstrapperApplication.cpp
```diff
@@ -807,15 +807,15 @@ class CWixStandardBootstrapperApplication : public CBalBaseBootstrapperApplicati
...
-            // Needs to match MsiEngineCalculateInstallUiLevel in msiengine.cpp in Burn.
-            BOOTSTRAPPER_ACTION_STATE packageAction = fExecute ? pPackage->executeAction : pPackage->rollbackAction;
-            fShowingInternalUiThisPackage = pPackage && pPackage->fDisplayInternalUI &&
-                                            BOOTSTRAPPER_ACTION_UNINSTALL != packageAction &&
-                                            BOOTSTRAPPER_ACTION_REPAIR != packageAction &&
-                                            (BOOTSTRAPPER_DISPLAY_FULL == m_command.display ||
-                                            BOOTSTRAPPER_DISPLAY_PASSIVE == m_command.display);
+                // Needs to match MsiEngineCalculateInstallUiLevel in msiengine.cpp in Burn.
+                BOOTSTRAPPER_ACTION_STATE packageAction = fExecute ? pPackage->executeAction : pPackage->rollbackAction;
+                fShowingInternalUiThisPackage = pPackage && pPackage->fDisplayInternalUI &&
```
rseanhall, line 813: This looks right, not sure how I made this mistake. You could do an optimization here since now we know `pPackage` is not `NULL`.

PR #536, opened 2021-07-19
Drop XP support for v3.14 => v4 transition.

barnson:
Drop XP support for v3.14 => v4 transition.

Already implemented in v4.


barnson:
Implements https://github.com/wixtoolset/issues/issues/6469.

PR #537, opened 2021-07-26
Correct sample for SetProperty/@Sequence

heaths:
Since the deferred custom action is scheduled only in `InstallExecuteSequence`, the `SetProperty/@Sequence` attribute must be set to "execute" since the default is "both". This leads to a LGHT0001 error about a missing `ActionRow` that can be difficult to diagnose.


heaths:
Where's the wix4 equivalent? I searched the org's repos to no avail.

PR #538, opened 2021-08-30
Add NativeMachine to Burn and WIX_NATIVE_MACHINE to UtilExtension

ericstj:
Add the ability for bundles and MSI's to detect when running on an ARM64 machine.

This directly plumbs the result of https://docs.microsoft.com/en-us/windows/win32/api/wow64apiset/nf-wow64apiset-iswow64process2 `pNativeMachine` argument, which is documented as 
> On success, returns a pointer to a possible [IMAGE_FILE_MACHINE_*](https://docs.microsoft.com/en-us/windows/win32/sysinfo/image-file-machine-constants) value identifying the native architecture of host system.

Issue: https://github.com/wixtoolset/issues/issues/6556
Wix4: https://github.com/wixtoolset/wix4/pull/37


src/chm/documents/bundle/bundle_built_in_variables.html.md
```diff
@@ -24,6 +24,7 @@ The Burn engine offers a set of commonly-used variables so you can use them with
 * LocalAppDataFolder - gets the well-known folder for CSIDL\_LOCAL\_APPDATA.
 * LogonUser - gets the current user name.
 * MyPicturesFolder - gets the well-known folder for CSIDL\_MYPICTURES.
+* NativeMachine - gets the [Image File Machine value](https://docs.microsoft.com/en-us/windows/win32/sysinfo/image-file-machine-constants) representing the native architecture of the machine. This property is only set on Windows 10, version 1511 (TH2) and higher.
```
rseanhall, line 27: I don't like the name `NativeMachine`, I think that only makes sense if you're looking at the `IsWow64Process2` documentation. Perhaps `NativeProcessorArchitecture`? Either way, probably should remove the word "native" from the documentation for `ProcessorArchitecture`.

ericstj: It is tied to “image file machine constants” as well https://docs.microsoft.com/en-us/windows/win32/sysinfo/image-file-machine-constants.  You can’t possibly use this property unless you read those docs due to the esoteric number values.

If we call it something different we risk making some sort of promise other than what the underlying API is making. I prefer to make this a direct pass through to that underlying API, naming and all.

ericstj: To be clear, I'm happy to change this to whatever folks decide, but wanted to share my reasoning for the current naming.  @barnson @robmen can you let me know what to do to establish consensus on this?  I'm happy to write something up if necessary.

barnson: I'm OK with a "pass-through" value so we don't have to make judgment calls about which platforms we should "promote." That said, when I made the original pass for ARM64/x64 support in Burn, I added both the Burn platform and native platform to `EngineRun`. (Though I did so using `GetNativeSystemInfo`. 💢) Having the platforms available as strings would make using them a lot prettier. I'd probably split the difference and say that v3 is good enough with inscrutable numbers and at some point in v4.0/v4.x, we should refactor the extant code to get strings in a new variable.

ericstj: Once you add string translations to this you make a promise that you'll update it once some new platform comes along and needs a string value.  If you just pass the integer through then folks can depend on whatever the Windows API returns, even if new values show up in the future.  IMHO that's better then the minor convenience of a string.  Folks will likely define wix variables or defines with expected integer values anyway.  What's your thinking about adding something like this to the extension to make that easier?
```XML
<!-- IMAGE_FILE_MACHINE_UNKNOWN 0 -->
<WixVariable Name"FileMachineUnknown" Value="0" />
<!-- IMAGE_FILE_MACHINE_I386 0x014c (332) -->
<WixVariable Name"FileMachineI386" Value="332" />
<!-- IMAGE_FILE_MACHINE_AMD64 0x8664 (34404) -->
<WixVariable Name"FileMachineAMD64" Value="34404" />
<!-- IMAGE_FILE_MACHINE_ARM64 0xAA64 (43602) -->
<WixVariable Name"FileMachineARM64" Value="43602" />
```
That way folks aren't blocked if a new platform comes along, and we could document these as part of the extension.

barnson: Providing "constants" is a decent workaround, though they'd have to be namespaced/reserved. OTOH, I'm not exactly worried about a new architecture just popping up out of nowhere.

ericstj: > I'm not exactly worried about a new architecture just popping up out of nowhere.

Look at this case.  We have 5 year old servicing branches that didn't anticipate this but needed to have installer work done to play nicely with it, without explicitly supporting it.  It would be nice if we didn't need to take toolchain updates to enable that.

barnson: > It would be nice if we didn't need to take toolchain updates to enable that.

Meh. It's not a bad goal but the probability of that happening is low. You might remember that .NET didn't have any ARM64 installers until after I did the work to support ARM64 in WiX. (You're totally welcome! No thanks needed!) That there's also a tweak required for x86-on-ARM64 is...unsurprising.

But to be clear, I'm not opposed to exposing the "raw" values.

ericstj: > It's not a bad goal but the probability of that happening is low

Agreed.

> You're totally welcome! No thanks needed!

Thank you indeed, I wasn't aware of that.  Been away from installer space for a bit, but I'm happy to come back to help.  This codebase brings back lots of fun memories.  What changes do you guys think are necessary to get this and the wix4 PR in?  Also I plan to attend Thursday's meeting if that's a better place to discuss.

barnson: https://www.youtube.com/watch?v=UPw-3e_pzqU

Thursday's meeting is probably where the discussion should happen, as it's a change in v3.x. I'll take a look at the v4 pull request before then.

src/chm/documents/customactions/osinfo.html.md
```diff
@@ -549,3 +549,17 @@ WixUtilExtension will automatically schedule the custom actions as needed after
...
+<h2>WixQueryNativeMachine properties</h2>
+
+<table>
+<tr>
+<td valign="top">
+<p>WIX_NATIVE_MACHINE</p>
+</td>
+
+<td>
+<p>Set to an [Image File Machine value](https://docs.microsoft.com/en-us/windows/win32/sysinfo/image-file-machine-constants) representing the native architecture of the machine. This property is only set on Windows 10, version 1511 (TH2) and higher.</p>
```
rseanhall, line 562: This is the only functional problem I see with the PR - the documentation says that the MSI property and Burn variable is only set on Windows 10, version 1511 and higher but they are always set. I vote for the implementation to match the documentation and not set the property or variable if the API isn't available.

I can see you were trying to follow the pattern from `WIX_WDDM_DRIVER_PRESENT` and `WIX_DWM_COMPOSITION_ENABLED`. If you look at [DetectWDDMDriver](https://github.com/wixtoolset/wix3/blob/58abd6993afba08b39e37b0e76b1790161df9231/src/ext/ca/wixca/dll/OsInfo.cpp#L382) and [DetectIsCompositionEnabled](https://github.com/wixtoolset/wix3/blob/58abd6993afba08b39e37b0e76b1790161df9231/src/ext/ca/wixca/dll/OsInfo.cpp#L420), if the corresponding API could not be dynamically loaded then it returns successfully without setting the property.

It also looks like you implemented `ProcNativeMachine` following the pattern of `ProcWow64`. But this conflicts with the pattern above because the caller can't tell whether the API was available or not. `ProcWow64` reports success and returns `FALSE` if the API wasn't available, which makes sense because there's no way it could be running on WOW in that case. Maybe `ProcNativeMachine` should return `E_NOTIMPL` or maybe there's no need to tell the difference between the API not being available and `IsWow64Process2` returning `IMAGE_FILE_MACHINE_UNKNOWN`, I don't know.

ericstj: Good catch.  You are right that today it will set IMAGE_FILE_MACHINE_UNKNOWN (0) rather than not setting the value when the API is not supported.

> maybe there's no need to tell the difference between the API not being available and IsWow64Process2 returning IMAGE_FILE_MACHINE_UNKNOWN, I don't know.

In our case we don't need to distinguish, but I could see someone wanting to know this.

I am fine with changing it to not set the property at all and make it consistent with the docs.

rseanhall:
Thanks!

@barnson I made some tweaks to the documentation and some style nits in https://github.com/wixtoolset/wix3/commit/38c8293f766f2708b2ffd4983abe9d930153c8a1.

@robmen Now that this is merged, the only things remaining for the v3.14 build is https://github.com/wixtoolset/wix3/pull/539 and https://github.com/wixtoolset/wix3/pull/540.

barnson:
👍

PR #540, opened 2021-09-10
Fix building bundles on FIPS-enabled machines

rseanhall:
Addresses complaints from https://github.com/wixtoolset/issues/issues/3992 and maybe https://github.com/wixtoolset/issues/issues/6428

### DIFF
--- a/src/tools/wix/Common.cs
+++ b/src/tools/wix/Common.cs
@@ -458,7 +458,7 @@ namespace Microsoft.Tools.WindowsInstallerXml
         internal static string GetFileHash(FileInfo fileInfo)
         {
             byte[] hashBytes;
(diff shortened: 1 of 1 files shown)


src/tools/wix/Common.cs
```diff
@@ -458,7 +458,7 @@ private static void RecursiveFileAttributes(string path, FileAttributes fileAttr
         internal static string GetFileHash(FileInfo fileInfo)
         {
             byte[] hashBytes;
-            using (SHA1Managed managed = new SHA1Managed())
+            using (SHA1 managed = new SHA1CryptoServiceProvider())
```
barnson, line 461: The `managed` name is bugging me. :) Otherwise, LGTM.

barnson:
LGTM

PR #545, opened 2022-08-28
Fix crash in thmviewer

rseanhall:


untra:
Looking forward to a new release of ThmViewer.exe that can help with design debugging. In v3 or v5!

